PR #1, opened 2021-04-01
Add checks for missing files and folders.

ddaspit:
Add checks to give clear messages if any of these files or folders are missing:

Experiment folder
Config file
Scripture file
Terms folder
Terms file

### DIFF
--- a/silnlp/common/corpus.py
+++ b/silnlp/common/corpus.py
@@ -123,7 +123,10 @@ def filter_parallel_corpus(corpus: pd.DataFrame, score_threshold: float) -> pd.D
 
 
 def get_scripture_path(iso: str, project: str) -> Path:
-    return MT_SCRIPTURE_DIR / f"{iso}-{project}.txt"
+    scripture_file = MT_SCRIPTURE_DIR / f"{iso}-{project}.txt"
+    if not scripture_file.exists():
+        raise RuntimeError(f"Can't find the Scripture file: '{scripture_file}'")
+    return scripture_file
 
 
 def include_books(corpus: pd.DataFrame, books: Set[int]) -> pd.DataFrame:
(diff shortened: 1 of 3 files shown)


silnlp/common/corpus.py
```diff
@@ -123,7 +123,10 @@ def filter_parallel_corpus(corpus: pd.DataFrame, score_threshold: float) -> pd.D
 
 
 def get_scripture_path(iso: str, project: str) -> Path:
-    return MT_SCRIPTURE_DIR / f"{iso}-{project}.txt"
+    scripture_file = MT_SCRIPTURE_DIR / f"{iso}-{project}.txt"
+    if not scripture_file.exists():
```
ddaspit, line 127: Rather than checking if the file exists in these functions, I think it would be better to add these checks where the file is actually opened. These functions only build the path. There might be valid cases where the you want to build the path for a non-existent file. The code that you added in the `load_config` function is a good example of what I would suggest.

silnlp/common/corpus.py
```diff
@@ -135,11 +138,17 @@ def exclude_books(corpus: pd.DataFrame, books: Set[int]) -> pd.DataFrame:
 
 
 def get_terms_metadata_path(list_name: str) -> Path:
-    return MT_TERMS_DIR / f"{list_name}-metadata.txt"
+    terms_dir = MT_TERMS_DIR / f"{list_name}-metadata.txt"
+    if not terms_dir.exists():
```
ddaspit, line 142: I would use `is_file()` instead of `exists()`. This method checks both that the path exists and that it is a file.

davidbaines: Yes, thanks that's much better.  I think there's an is_dir()  for folders too.

silnlp/common/corpus.py
davidbaines: Only checking when we try to open the file is causing me delays and is the main reason I'd like to put in the checks. In particular if the target data is missing I have to wait until the source vocabularies have been built before I find out.  It might be just a typo in the config but it happens often enough.  

An alternative could be a separate script or preprocess option that uses the current logic to find and report on the existence of all the files mentioned in a config file.

ddaspit: You could still check if the file exists earlier. I just wouldn't check in these functions. It sounds like you want to check if all of the data files exist before preprocessing. I would suggest that perform the check at the beginning of the `Config.preprocess()` method in `config.py`. In that method, you could iterate through `src_file_paths` and `trg_file_paths` and check if each file exists. If a file doesn't exist, you would raise an exception.

johnml1135:
What is the status of this?  Should it be merged? Should it be killed?

johnml1135:
This is related to #14 - I will try to separately implement it doing Damien's recommendation - check at the beginning of Preprocess.

ddaspit:
I am going to go ahead and close this PR, since @johnml1135 will implement this in a different PR.

Issue #7, opened 2021-04-05
ClearML build using docker


johnml1135:
Should we use docker for the ClearML jobs?  More generally, would it be helpful to use docker for development?  We could then host it in other locations and remote in using VSC...

Some references for poetry and docker working together:
https://github.com/python-poetry/poetry/discussions/1879
https://stackoverflow.com/questions/53835198/integrating-python-poetry-with-docker

johnml1135:
Or, use poetry virtualenv - ClearML needs to be one or the other.  Would both together be overkill?

johnml1135:
Step1: cpu only - using:
```
FROM mcr.microsoft.com/dotnet/runtime-deps:5.0.5-focal-amd64
RUN apt-get update -y && apt-get install python3.8 -y
```
Next,  Push that to docker hub.  This is our image.
Use that image for running our code.
Next.  Update the code to run  `dotnet tool restore` before everything - to always pull the right stuff.
ClearML should see the poetry file and pull everything automatically (hopefully).

johnml1135:
Steps to done:
(cpu-only)
* Convert Path objects to S3Path
* Run silnlp.common.extract_corpora (uses machine)
* copy configuration from another experiment, edit.  
* Run silnlp.nmt.preprocess (uses machine)
(gpu)
* edit configuration - maxsteps = 1000
* Run silnlp.nml.train (uses gpu)
(logging results to compare)
* Reroute csv blue scores to ClearML in sil.nmt.test

johnml1135:
What to do about:
* meteor_path = Path(os.getenv("METEOR_PATH", "."), "meteor-1.5.jar")
* plugin_file_path=Path(os.getenv("BETA_INV_PLUGIN_PATH", ".")),
* mgiza_path = Path(os.getenv("MGIZA_PATH", "."), "mgiza")
* fast_align_path = Path(os.getenv("FAST_ALIGN_PATH", "."), "fast_align")

Are these extra things that should be installed into the docker image?

johnml1135:
Scrap everything and start again - machine needs paths.  We need to mount it.

I looked for a solution and documented what I found here: https://stackoverflow.com/questions/67496760/mounting-an-s3-bucket-in-docker-in-a-clearml-agent/67497264#67497264

Only a small number of changes were needed - but I also added the files for mounting rclone and the clearml-agent.

johnml1135:
Everything is done and integrated into master.  We can open a new issue if we need to.

PR #21, opened 2021-05-14
Clearml and Google Drive

johnml1135:
…other functions (preprocess, train and test).


johnml1135:
We can merge now to try out - I have not tested other functions.  They will need to have clearml code added to them as well.  This could be done by anybody though and the lifting of getting S3 working is done.  I have S3 mounted on my PC, but we also need to work and get it mounted on the other drives as well.

There is a significant possibility that S3 will not like being treated as a folder structure (appends, other trickery) and may crash and need to be restarted.  If this becomes a bigger issue, we may need to investigate other solutions.

I documented my work trying to resolve this issue here: https://stackoverflow.com/questions/67496760/mounting-an-s3-bucket-in-docker-in-a-clearml-agent/67497264#67497264

johnml1135:
This is a fix for #7 issue.

scripts/clearml_agent/mount_aqua_on_data.bat
```diff
@@ -0,0 +1,9 @@
+rem Download rclone from: https://rclone.org/
+rem Extract into a folder.
+rem Copy this file and rclone.conf to the new folder
+rem You will also need the api key and secret installed through awscli for this to work.
+rem Note that you can map \\wsl$\Ubuntu to a drive letter and then browse the data.
+call wsl.exe fusermount -uz /data/aqua-ml-data
```
ddaspit, line 6: What about Linux? Also, maybe this path should be specified by the `SIL_NLP_DATA_PATH` environment variable.

silnlp/common/environment.py
```diff
@@ -12,6 +12,9 @@ def get_data_dir() -> Path:
     sil_nlp_data_path = os.getenv("SIL_NLP_DATA_PATH")
     if sil_nlp_data_path is not None:
         return Path(sil_nlp_data_path)
+    auqa_ml_path = Path("/data/aqua-ml-data")
```
ddaspit, line 15: rclone should be mounted to the path specified by `SIL_NLP_DATA_PATH`, then this code won't be necessary.

silnlp/common/extract_corpora.py
```diff
@@ -26,6 +29,10 @@ def main() -> None:
 
     projects: Set[str] = set(args.projects)
 
+    task = Task.init(
```
ddaspit, line 32: I guess this is mostly done as a test ClearML task. I don't know how useful this would be in real-world use.

ddaspit (changes requested):

johnml1135:
Have ClearML be an option in Extract Corpora
Make a nmt/experiment.py file - base experiment (no clearml) class for preprocess, train and test.  Child class with ClearML.

johnml1135:
Run with 3 steps (create checkpoint there) - without GPU.

johnml1135:
Damien, are you OK merging this in now?

johnml1135:
Needs one more commit after clearml 1.0.3 comes out to update the lock file.

ddaspit:
I didn't get a chance to review this until today. The changes look good to me.

ddaspit:
Feel free to delete the branch if you don't need it anymore.

PR #54, opened 2021-06-23
Initial integration tests

johnml1135:
Initial integration tests (extract_corpus and preprocess)
Create assets folder


tests/test_extract_corpora.py
```diff
@@ -0,0 +1,32 @@
...
+
+
+pp_dir = Path(__file__).parent / "Paratext/projects"
+
+scr_dir = Path(__file__).parent / "MT/scripture"
+scr_truth_dir = Path(__file__).parent / "MT/scripture_truth"
+
+
+def test_extract_corpora():
+    shutil.rmtree(scr_dir, ignore_errors=True)
```
ddaspit, line 15: I think it would be good to use `tempfile.TemporaryDirectory` here and set `MT_SCRIPTURE_DIR` to the temp dir path. Maybe something like this:

```
with tempfile.TemporaryDirectory() as src_dir:
    MT_SCRIPTURE_DIR = Path(src_dir)
```

I think this would be safer.

tests/test_preprocess.py
```diff
@@ -0,0 +1,30 @@
...
+exp_subdirs = [folder for folder in exp_truth_dir.glob("*/")]
+
+
+@pytest.mark.parametrize("exp_folder", exp_subdirs)
+def test_preprocess(exp_folder):
+    exp_truth_path = os.path.join(exp_truth_dir, exp_folder)
+    config_file = os.path.join(exp_truth_path, "config.yml")
+    assert os.path.isfile(config_file), "The configuraiton file config.yml does not exist for " + exp_folder.name
+    experiment_path = exp_dir / exp_folder.name
+    shutil.rmtree(experiment_path, ignore_errors=True)
```
ddaspit, line 21: See my suggestion above regarding the use of `tempfile.TemporaryDirectory`.

ddaspit (changes requested):

tests/test_preprocess.py
johnml1135: I tried your suggestion - the main issue is that MT_SCRIPTURE_DIR is pre-set in the extract corpora file and is not able to be modified.  To do this, the code would have to be refractored in some way.  One way could be to create a class of "Environment" that is passed into these various functions.  This would involve around 90 modifications.  It is possible - let me try.

johnml1135: So, I updated environment to be able to be modifiable (which was needed).  There is still an issue with temp directory.  Often, after a test is run, I want to look at the artifacts to see what changed.  It is much easier to see that in a predictable folder.  I made a "temp_experiments" and "temp_scripture" folder.  If there is a better way of doing it, let me know.

tests/helper.py
```diff
@@ -0,0 +1,73 @@
...
+    LOGGER = logging.getLogger()
+    LOGGER.setLevel(logging.INFO)
+    LOGGER.handlers = []  # removes all handlers
+    formatter = logging.Formatter("%(asctime)s - %(name)s - %(levelname)s - %(message)s")
+    fh = logging.FileHandler(os.path.join(exp_folder, "log.txt"))
+    fh.setFormatter(formatter)
+    LOGGER.addHandler(fh)
+
+
+def remove_previously_created_files(folder: str, extensions_to_delete=("txt", "json")):
```
ddaspit, line 16: This doesn't seem to be used anywhere.

tests/helper.py
```diff
@@ -0,0 +1,73 @@
...
+        elif tf.endswith((".vocab")):
+            # make sure it's not too different.
+            tf_set = set([l.split()[0] for l in open(tfp, "r", encoding="utf-8").readlines()])
+            cf_set = set([l.split()[0] for l in open(cfp, "r", encoding="utf-8").readlines()])
+            diff_set = (tf_set - cf_set) | (cf_set - tf_set)
+            # Sentencepiece may not always make it excaclty the same.  If there is a big change, flag it.
+            assert (
+                len(diff_set) < len(tf_set) * 0.01
+            ), f"There are {len(diff_set)} differences in the file {tf}.  They are: {diff_set}"
+        elif tf.endswith((".model")):
```
ddaspit, line 67: If there is no good way to compare these, then I would remove these files from the repo. I don't really like having binary files in the repo anyway.

tests/helper.py
```diff
@@ -0,0 +1,73 @@
...
+                assert (
+                    tf_content[i].strip() == cf_content[i].strip()
+                ), f"line {i} in {tf} should be:\n  {tf_content[i]}\nbut is:\n  {cf_content[i]}"
+        elif tf.endswith((".vocab")):
+            # make sure it's not too different.
+            tf_set = set([l.split()[0] for l in open(tfp, "r", encoding="utf-8").readlines()])
+            cf_set = set([l.split()[0] for l in open(cfp, "r", encoding="utf-8").readlines()])
+            diff_set = (tf_set - cf_set) | (cf_set - tf_set)
+            # Sentencepiece may not always make it excaclty the same.  If there is a big change, flag it.
+            assert (
```
ddaspit, line 64: This makes me a bit nervous. I like tests to be stable and deterministic. This is probably happens because we do not set the seed for sentencepiece. You could try calling `set_random_generator_seed` on the sentencepiece module in the tests to see if this makes sentencepiece deterministic.

tests/test_preprocess.py
```diff
@@ -0,0 +1,32 @@
...
+SNE._MT_EXPERIMENTS_DIR.mkdir(exist_ok=True)
+exp_truth_dir = SNE._MT_DIR / "Experiments"
+exp_subdirs = [folder for folder in exp_truth_dir.glob("*/")]
+
+
+@pytest.mark.parametrize("exp_folder", exp_subdirs)
+def test_preprocess(exp_folder):
+    exp_truth_path = os.path.join(exp_truth_dir, exp_folder)
+    config_file = os.path.join(exp_truth_path, "config.yml")
+    assert os.path.isfile(config_file), "The configuraiton file config.yml does not exist for " + exp_folder.name
```
ddaspit, line 21: "configuration" is misspelled.

tests/test_preprocess.py
ddaspit: Can you not pass the temp directory path to `compare_folders`?

ddaspit (changes requested):

tests/test_preprocess.py
johnml1135: I can, but often if there is something "wrong" or "different", we will want to look at it in Beyond Compare, etc.  if it is a temp folder, it would be challenging to find the artifacts made to do the comparison (as happens often).  I am fine putting it anywhere as long as it is deterministic and named.

silnlp/assets/Ref/CommentTags.xml.BAK
```diff
@@ -0,0 +1,5 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
```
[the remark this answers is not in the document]
johnml1135: Didn't catch those files - removing now.

tests/helper.py
johnml1135: I agree - I pulled it over from another project I was working on and didn't realize it is not needed for this one.  Removed.

johnml1135: That is, removed remove_previously_created_files(...)

tests/helper.py
johnml1135: I am comparing for file size difference.  I want to make sure that something similar was made.  I think it is a non-trivial comparison (that it was created and that the file is similarly sized).  All the preprocess outputs are hard to compare, vocal, etc.  They are not always created the same way (extra word parts, different order).  This is my best shot at a meaningful comparison.  I am open to different ideas.

tests/helper.py
johnml1135: Set the seed for sentencepiece: https://github.com/sillsdev/silnlp/blob/9184429981abf98cf3b108da5dd93732b754739a/silnlp/nmt/config.py#L322-L326

also needed to sort the filenames for SentencePiece to make .model deterministic:
https://github.com/sillsdev/silnlp/blob/9184429981abf98cf3b108da5dd93732b754739a/silnlp/nmt/config.py#L125-L126

tests/test_preprocess.py
johnml1135: fixed.

johnml1135:
So the tests are failing - it is now always looking for "corpus pairs" even though some configurations don't have them.  It should act gracefully if there are no corpus pairs.  What is the desired behavior?

Error:
```
================================== FAILURES ===================================
________________________ test_preprocess[exp_folder0] _________________________

exp_folder = WindowsPath('c:/Users/johnm/Documents/repos/silnlp/tests/MT/Experiments/baseline-child-model')

    @pytest.mark.parametrize("exp_folder", exp_subdirs)
    def test_preprocess(exp_folder):
...   
>       config = load_config(experiment_path)

tests\test_preprocess.py:28: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
silnlp\nmt\config.py:1666: in load_config
    return Config(exp_dir, config)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <silnlp.nmt.config.Config object at 0x0000025697F5E9D0>
exp_dir = WindowsPath('c:/Users/johnm/Documents/repos/silnlp/tests/MT/temp_experiments/baseline-child-model')
config = {'data': {'character_coverage': 1.0, 'eval_features_file': 'c:\\Users\\johnm\\Documents\\repos\\silnlp\\tests\\MT\\tem...'model_dir': 'c:\\Users\\johnm\\Documents\\repos\\silnlp\\tests\\MT\\temp_experiments\\baseline-child-model\\run', ...}

    def __init__(self, exp_dir: Path, config: dict) -> None:
...    
        self.exp_dir = exp_dir
        self.root = config
    
>       self.corpus_pairs = parse_corpus_pairs(data_config["corpus_pairs"])
E       KeyError: 'corpus_pairs'
 ```

File: silnlp/nmt/config.py [hunk 8]
```diff
@@ -642,6 +645,7 @@ def has_parent(self) -> bool:
     def set_seed(self) -> None:
         seed = self.data["seed"]
         set_seed(seed)
+        sp.set_random_generator_seed(seed)
```
ddaspit, line 648: This won't give the expected behavior. You can only set it once for a process and it must be set before you use sentencepiece anywhere else (see sentencepiece [issue](https://github.com/google/sentencepiece/issues/635)). I would just set this for the tests.

ddaspit (changes requested):
Please remove the ".model" files. I really don't want to have binary files in the repo. Checking the sentencepiece vocabulary files should be sufficient.

ddaspit:
The `langs` config is no longer supported. All config files must use `corpus_pairs` now. The config files for the tests should be updated to use `corpus_pairs`.

tests/helper.py
```diff
@@ -0,0 +1,46 @@
...
+
+
+def compare_folders(truth_folder: str, computed_folder: str):
+    truth_files = os.listdir(truth_folder)
+    for tf in truth_files:
+        tfp = os.path.join(truth_folder, tf)
+        cfp = os.path.join(computed_folder, tf)
+        assert os.path.isfile(cfp), "The file " + tf + " should have been but was not created."
+        if tf == "log.txt":
+            tf_content = open(tfp, "r", encoding="utf-8").readlines()
```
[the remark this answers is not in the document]
johnml1135: changed to with open(...) as f:

silnlp/nmt/config.py
johnml1135: Fixed - called for each preprocessing test.

johnml1135:
Should have fixed all issues including removing .model files.

ddaspit (commented):
You missed a couple of comments I made involving `SilNlpEnv` in `environment.py`. Everything else looks good.

silnlp/common/environment.py
```diff
@@ -1,48 +1,99 @@
...
+            return gutenberg_path
+
+        data_root = self._ROOT_DIR / "data"
+        LOGGER.info(
+            f"Using workspace: {data_root}.  To change the workspace, set the environment variable SIL_NLP_DATA_PATH."
+        )
+        return data_root
+
+
+SNE = SilNlpEnv()
```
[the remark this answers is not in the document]
johnml1135: SIL_NLP_ENV?

silnlp/common/environment.py
```diff
@@ -1,48 +1,99 @@
...
 
-def get_data_dir() -> Path:
-    sil_nlp_data_path = os.getenv("SIL_NLP_DATA_PATH")
-    if sil_nlp_data_path is not None:
-        temp_path = Path(sil_nlp_data_path)
-        if temp_path.is_dir():
-            return Path(sil_nlp_data_path)
+class SilNlpEnv:
+    def __init__(self):
+        self._ROOT_DIR = Path.home() / ".silnlp"
```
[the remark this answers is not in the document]
johnml1135: lowercase.

johnml1135:
All issues should be resolved now.

ddaspit (approved):
This looks good. I see that `corpus_pairs_config.py` and `langs_config.py` are in this PR. I'm not sure if the current master has been merged in yet. Just make sure that these files don't get committed back to master when this PR is merged.

johnml1135:
Okay, I'll make sure.

PR #82, opened 2021-10-04
Fix for copying parent "best" checkpoint.

johnml1135:


johnml1135:
@ddaspit and @mmartin9684-sil , this should fix the issue you were having with parent projects.

Issue #81, opened 2021-10-04
Add Test to clearml experiment


johnml1135:
Add the test step.  Fix any S3 bucket things.
Also, log blue scores throughout.

johnml1135:
From Micheal:
* Everything but meteor
* Best and last
* Log for plotting in ClearML
* Plot incremental blue scores

johnml1135:
Can we use tensorboard?
* https://github.com/allegroai/clearml/blob/master/examples/frameworks/tensorflow/tensorboard_pr_curve.py

PR #86, opened 2021-10-12
#81 add testing to clearml

johnml1135:
Tested here: https://app.pro.clear.ml/projects/d6d62edafd4b42d393310cd3f97bde1d/experiments/3493d23c676745399c9afd2b418d79bf/info-output/log?columns=selected&columns=type&columns=name&columns=tags&columns=status&columns=project.name&columns=users&columns=started&columns=last_update&columns=last_iteration&columns=parent.name&order=-last_update


silnlp/nmt/config.py
```diff
@@ -407,17 +407,20 @@ def build_vocab(
 
 
 def get_checkpoint_path(model_dir: Path, checkpoint_type: CheckpointType) -> Tuple[Optional[Path], Optional[int]]:
+    model_dir = SIL_NLP_ENV.get_source_experiment_path(model_dir)
+    ckpt_and_step = (None, None)
```
ddaspit, line 411: I think it would be more clear if these were two separate variables.

silnlp/nmt/experiment.py
```diff
@@ -3,6 +3,7 @@
 import os
 from dataclasses import dataclass
 from silnlp.common.environment import SIL_NLP_ENV
+from silnlp.nmt.test import test
```
ddaspit, line 6: Can you make this a relative import so that it is consistent with the rest of the codebase?

Issue #78, opened 2021-09-20
Bible Machine Alignments


johnml1135:
This issue is to track the progress to making generally available 100's if not 1000's of Bible translations with machine alignments.  This has a few audiences:
* multi-corpora offline analysis 
* Seeds for ML work
* Starting point for alignment to be used for enhanced resources
All Bibles will be aligned to greek and hebrew, as is in the paratext projects.

Brainstorming implementation:
* Process all Bibles as per a simple json file in the format: 
  * {alignment:fast-align { en-ulb: "all", es-ulb: "hi-ulb, fr-ulb"} }
* Word by word alignment.
* Store this json file and alignment_records.csv in a githib repo.
  * the README.md has a description of the content as well as a massive table of links to all the alignments and overall alignment scores
  * The alignments are stored in a public S3 bucket (just like Door43)
* Create alignment_records.csv that contains:
  * Source and target languages
  * Alignment method
  * overall alignment score
  * timestamp
  * S3 bucket link for alignment per verse
  * S3 bucket link for alignment scores per verse
* This data will normally be created on a local PC

johnml1135:
- [ ] Identify source bibles
- [ ] Identify repo to store extracted Bibles
- [ ] Identify repo to store alignments
- [ ] setup S3 bucket (read-only public)
- [ ] Generate a script to make 1 alignment as we want it to be made - review it to ensure it's suitability
- [ ] Generate script in this repo to create all alignments as specified by the json file (that is, if they are not already made)
- [ ] Generate 10 or so alignments and put it on GitHub - verify everything is working as expected including:
  - [ ] licenses
  - [ ] can upload to the S3 bucket
  - [ ] can download from the S3 bucket
  - [ ] only new files are created
- [ ] Develop initial list of Bible pairs for the repo
- [ ] Update the json file, run the script, upload to GitHub.
- [ ] Release to the public

johnml1135:
First steps:
* Align a set (10?) Bibles to greek and Hebrew with IBM-4
  * Which Greek/Hebrew corpuses can we use?  Use whatever Clear is using (so we can use Clear Engine) (Freely Licensed):
    * Nestle1904
    * SBLGNT (greek new testement)
    * Wesminster Leningrad Codex (WLC)
  * different versification between gr/hr and english.   Currently uses English versification
* Then selections as needed
* Start with word level and then assess - most tools use word tokenized alignments

Audience 1:
[...]

PR #111, opened 2021-12-03
#78 machine alignments fix

johnml1135:
This is a fix on the previous pull request.


johnml1135:
@ddaspit - I think this is the proper fix.  Can you give it a once over to see if I did everything correctly?

Issue #113, opened 2021-12-08
Normalize language codes to ISO-639-3 when creating extract files


mmartin9684-sil:
When extracting verse text from a Paratext project into an extract file, the file name is uses the ISO language code read from the project's Settings.xml file.  These values vary between ISO-639-2 and ISO-639-3 language codes, making it more difficult to easily identify extract files in the same language.  Normalizing the ISO language code from the settings file (to ISO-639-3) when assigning the file name would make the extract files easier to use.

ddaspit:
@mmartin9684-sil  According to [BCP 47](https://tools.ietf.org/search/bcp47), language tags use the ISO 639-1 code (two letter code) if there is one, otherwise they use the ISO 639-3 code (three letter code). So the question is "what do we want the code in the file name to represent?" If we want it to represent a BCP 47 language tag, then the current behavior is correct. If we want it to represent an ISO 639-3 code, then we should change it. I took a look at Paratext and the `LanguageIsoCode` element in the PT settings is intended to represent a BCP 47 code. We are currently parsing only the language subtag from the full tag. We might actually want to include the full BCP 47 tag in the file name.

PR #114, opened 2021-12-09
Normalize ISO code from Settings.xml when creating the extract file n…

mmartin9684-sil:
...ame.


mmartin9684-sil:
This should address issue #113.

Issue #137, opened 2022-07-15
Improvements to split_parallel.py - currently it includes blank lines in test and validation sets.


davidbaines:
The split parallel code could be improved in a couple of ways:
Check whether either the source or target line is blank and if so do not use that pair for either the test or validation sets.
Check whether the source and target lines are identical and if so do not use that pair for either the test or validation sets.

PR #138, opened 2022-07-15
#137 Fix issue 

davidbaines:
Fix split_parallel.py so that test and validation sets don't contain blank or identical lines.


mmartin9684-sil (approved):
Looks good.

If we want this tool to be used for splitting Scripture pairs, we'll need it to handle "<range>" lines as well.

PR #141, opened 2022-10-26
Extractions and Paratext directory from environment variable

vesaakerman:
Make it possible to change the location of extractions and Paratext files


davidbaines (commented):
Thanks Vesa, These changes would make our eBible work easier.

ddaspit (approved):

PR #191, opened 2023-08-22
Fix issue where updated tokenizer not used during inference, add segment length stats to preprocess

mshannon-sil:
When I refactored the tokenizer code during the previous request, there was an unintended consequence where inferencing would unsuccessfully try to update the tokenizer even though the tokenizer had already been updated during preprocessing. This update adds an optional argument to get_tokenizer() to differentiate between preprocessing and other steps, so that during any step other than preprocessing it will just use the tokenizer found in the project directory.

I also added functionality to track the following stats: num segment lengths > 200, max segment length, and avg segment length. These will be helpful for tokenizer experiments.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/191)

### DIFF
--- a/silnlp/nmt/config.py
+++ b/silnlp/nmt/config.py
@@ -520,7 +520,7 @@ class Config(ABC):
                         for line in f:
                             src_data.append(len(line.split()))
                 stats_file.write(f"Src segment length statistics:\n")
-                stats_file.write(f"Num seg lengths > 200: {sum(seg_length > 200 for seg_length in src_data)}\n")
+                stats_file.write(f"Num seg lengths >= 200: {sum(seg_length >= 200 for seg_length in src_data)}\n")
                 stats_file.write(f"Max seg length: {max(src_data)}\n")
                 stats_file.write(f"Avg seg length: {sum(src_data)/len(src_data)}\n")
                 trg_data = []
@@ -529,7 +529,7 @@ class Config(ABC):
                         for line in f:
                             trg_data.append(len(line.split()))
                 stats_file.write(f"Trg segment length statistics:\n")
-                stats_file.write(f"Num seg lengths > 200: {sum(seg_length > 200 for seg_length in trg_data)}\n")
+                stats_file.write(f"Num seg lengths >= 200: {sum(seg_length >= 200 for seg_length in trg_data)}\n")
                 stats_file.write(f"Max seg length: {max(trg_data)}\n")
                 stats_file.write(f"Avg seg length: {sum(trg_data)/len(trg_data)}\n")
         return train_count
--- a/silnlp/nmt/config.py
+++ b/silnlp/nmt/config.py
@@ -510,6 +510,28 @@ class Config(ABC):
                 train_count += self._write_scripture_data_sets(tokenizer, pair, stats)
             else:
                 train_count += self._write_basic_data_sets(tokenizer, pair)
+        if stats:    
+            with ExitStack() as stack:
+                stats_file: Optional[TextIO] = None
+                stats_file = stack.enter_context(self._open_append("tokenization_stats.txt"))
+                src_data = []
+                for src_tok_file in self.exp_dir.glob("*.src.txt"):
+                    with open(self.exp_dir / src_tok_file, "r+", encoding="utf-8") as f:
+                        for line in f:
+                            src_data.append(len(line.split()))
+                stats_file.write(f"Src segment length statistics:\n")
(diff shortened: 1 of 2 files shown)


ddaspit (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/191)* status: all files reviewed, 2 unresolved discussions (waiting on @mshannon-sil)

___
*[`silnlp/nmt/hugging_face_config.py` line 8 at r1](https://reviewable.io/reviews/sillsdev/silnlp/191#-NcyB45J6UbekGE_KBo3:-NcyB45J6UbekGE_KBo4:bjd8jh2) ([raw file](https://github.com/sillsdev/silnlp/blob/1f872bd6984eb095f72cbdc11cd1dac0b676be59/silnlp/nmt/hugging_face_config.py#L8)):*
> ```Python
> from itertools import repeat
> from pathlib import Path
> from typing import Any, Dict, Iterable, List, Optional, Set, TextIO, Tuple, TypeVar, Union, cast
> ```

I don't think this import is needed.
___
*[`silnlp/nmt/hugging_face_config.py` line 393 at r1](https://reviewable.io/reviews/sillsdev/silnlp/191#-NcyA_yt7vopSvB88eIH:-NcyA_yt7vopSvB88eII:bt45266) ([raw file](https://github.com/sillsdev/silnlp/blob/1f872bd6984eb095f72cbdc11cd1dac0b676be59/silnlp/nmt/hugging_face_config.py#L393)):*
> ```Python
>     def get_tokenizer(self, preprocess: bool=False) -> PreTrainedTokenizer:
>         if self._tokenizer is None:
>             if not preprocess:
> ```

This feels like two different methods: `get_tokenizer` and `get_or_create_tokenizer`. I think it would be better to refactor this single method into two methods.

mshannon-sil (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/191)* status: all files reviewed, 2 unresolved discussions (waiting on @ddaspit)

___
*[`silnlp/nmt/hugging_face_config.py` line 8 at r1](https://reviewable.io/reviews/sillsdev/silnlp/191#-NcyB45J6UbekGE_KBo3:-Nd1NxT14mwHob-kkg9O:b-mgjrkf) ([raw file](https://github.com/sillsdev/silnlp/blob/1f872bd6984eb095f72cbdc11cd1dac0b676be59/silnlp/nmt/hugging_face_config.py#L8)):*

I don't think this import is needed.
</blockquote></details>

TextIO is being used in hugging_face_config.py line 384: `stats_file: Optional[TextIO] = None`, so I think it's better to keep it, although when the import was missing previously it didn't seem to be causing any errors.

mshannon-sil (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/191)* status: all files reviewed, 2 unresolved discussions (waiting on @ddaspit)

___
*[`silnlp/nmt/hugging_face_config.py` line 393 at r1](https://reviewable.io/reviews/sillsdev/silnlp/191#-NcyA_yt7vopSvB88eIH:-Nd1chtjCwP7euEEdKFG:b-896fix) ([raw file](https://github.com/sillsdev/silnlp/blob/1f872bd6984eb095f72cbdc11cd1dac0b676be59/silnlp/nmt/hugging_face_config.py#L393)):*

This feels like two different methods: `get_tokenizer` and `get_or_create_tokenizer`. I think it would be better to refactor this single method into two methods.
</blockquote></details>

Done.

ddaspit (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/191#-:-Nd2JkIhCEnJKTmAixkR:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @mshannon-sil)

___
*[`silnlp/nmt/hugging_face_config.py` line 8 at r1](https://reviewable.io/reviews/sillsdev/silnlp/191#-NcyB45J6UbekGE_KBo3:-Nd2JbOP3Q-7aZpptRe_:b-d95zgv) ([raw file](https://github.com/sillsdev/silnlp/blob/1f872bd6984eb095f72cbdc11cd1dac0b676be59/silnlp/nmt/hugging_face_config.py#L8)):*

TextIO is being used in hugging_face_config.py line 384: `stats_file: Optional[TextIO] = None`, so I think it's better to keep it, although when the import was missing previously it didn't seem to be causing any errors.
</blockquote></details>

That makes sense.

PR #199, opened 2023-09-19
Normalize before calculating missing characters and trained tokens

mshannon-sil:
I added some code to apply normalization before calculating missing characters and trained tokens.

For characters I simply pass the whole set of characters from the dataset into `normalize_target_all()` and strip out any whitespace, removing any resulting empty strings, before calculating missing characters.

For trained tokens, I created a new `normalize()` function inside of the `HuggingFaceTokenizer` class. I instantiate an instance of this class and use it to initialize a custom Normalizer, which I assign to the sentencepiece tokenizer that will be trained. After it trains, I change the sentencepiece tokenizer back to using a non-custom normalizer because the model can't save with a custom one, and that would be a problem since the saved json file is needed to access its merge rules.

I also stopped adding the `lang_code` to the sentencepiece vocab in the json file, so it's only being added to the NLLB tokenizer using `add_special_tokens()` just like how it was before all the changes were made.

In addition, I made a few minor changes. I removed the `_add_merges()` function that was leftover from a previous implementation of the trained tokens, and I tweaked the early stopping condition so that it works when `early_stopping` is set to false in the config.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/199)

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -36,20 +36,15 @@
 		"vscode": {
 			// Set *default* container specific settings.json values on container create.
 			"settings": {
-				"python.defaultInterpreterPath": "/usr/local/bin/python",
-				"python.linting.enabled": true,
-				"python.formatting.autopep8Path": "/usr/local/py-utils/bin/autopep8",
-				"python.formatting.blackPath": "/usr/local/py-utils/bin/black",
-				"python.formatting.yapfPath": "/usr/local/py-utils/bin/yapf",
-				"python.linting.banditPath": "/usr/local/py-utils/bin/bandit",
-				"python.linting.flake8Path": "/usr/local/py-utils/bin/flake8",
-				"python.linting.mypyPath": "/usr/local/py-utils/bin/mypy",
-				"python.linting.pycodestylePath": "/usr/local/py-utils/bin/pycodestyle",
-				"python.linting.pydocstylePath": "/usr/local/py-utils/bin/pydocstyle",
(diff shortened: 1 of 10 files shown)


mshannon-sil:
@ddaspit Just pushed changes to fix the formatting with Black, Flake8, Isort, Pylint, and Pylance.

ddaspit (changes requested):
> Reviewed 10 of 10 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/199)* status: all files reviewed, 1 unresolved discussion (waiting on @mshannon-sil)

___
*[`silnlp/nmt/tokenizer.py` line 30 at r1](https://reviewable.io/reviews/sillsdev/silnlp/199#-NexYXJjAo3oKwyaObfO:-NexYXJjAo3oKwyaObfP:bvtq202) ([raw file](https://github.com/sillsdev/silnlp/blob/114501a6012589feec4b99ac724980733013c86d/silnlp/nmt/tokenizer.py#L30)):*
> ```Python
> 
>     @abstractmethod
>     def normalize(self, line: NormalizedString) -> None:
> ```

I don't think this needs to be part of the `Tokenizer` interface, since it is specific to Huggingface.

ddaspit (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/199#-:-NeyQ77T5S5VV0n6Cw54:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @mshannon-sil)

PR #233, opened 2023-11-28
Add ability to train models and do inference on specific chapters

isaac091:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/233)


davidbaines:
Hi Isaac,

This will be a very useful addition to our code. Thanks for your work on it.

Once this is reviewed and in the master branch, how should I specify the
chapters that are to be included or excluded - do they go on the command
line, or in the config file?
Also is there a strict format for specifying the chapters or are there more
than one formats that would work?

All the best,
David

isaac091:
Hi David,

Thanks for the questions. You will be able to specify chapters for training with the preexisting corpus_books and test_books arguments in the config file, and likewise you will be able to use the --books command line argument to specify chapters for inferencing. I added a [section](https://github.com/sillsdev/silnlp/wiki/Configure-a-model#alternative-syntax-for-corpus_books-and-test_books-to-use-chapter-specification-book-ranges-and-subtraction) to the wiki describing the new syntax, but the old syntax of comma-separated lists of books will still work as well.

davidbaines:
Hi Isaac,

Oh, that's great.  Thank you for adding documentation - it makes all
the difference. There are so many options, way too many to remember. We
don't have them all documented. Documenting new features as they are
included will be a very helpful practice.
Thank you!

All the best,
David

ddaspit (changes requested):
> Reviewed 5 of 5 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/233)* status: all files reviewed, 2 unresolved discussions (waiting on @isaac091)

___
*[`silnlp/nmt/translate.py` line 66 at r1](https://reviewable.io/reviews/sillsdev/silnlp/233#-NkRXSCn0lTlJnaCrpt5:-NkRXSCn0lTlJnaCrpt6:b-wqfuhb) ([raw file](https://github.com/sillsdev/silnlp/blob/d815a99b2ad27b6873f5cf36d6931ca8a9c37837/silnlp/nmt/translate.py#L66)):*
> ```Python
>             return
> 
>         if any(len(book_nums[book]) > 0 for book in book_nums) and trg_project is not None:
> ```

It would be good to add `trg_projects` property to `Config`. You can use it to set the default target project if one is not specified, similar to how the default source project is specified.
___
*[`silnlp/nmt/translate.py` line 90 at r1](https://reviewable.io/reviews/sillsdev/silnlp/233#-NkRZ923AFgK1ACXXiVk:-NkRZ923AFgK1ACXXiVl:b-8t2krs) ([raw file](https://github.com/sillsdev/silnlp/blob/d815a99b2ad27b6873f5cf36d6931ca8a9c37837/silnlp/nmt/translate.py#L90)):*
> ```Python
>                         src_project,
>                         book,
>                         output_path,
> ```

In this case, we want to save the book to a folder that matches the target project name.

isaac091:
Currently, translations not in a target project folder follow the structure of the source project, and translations in a target project folder follow the structure of the target project and missing/not translated chapters are filled in by the target project.

Since we are now using a target project from the config as a default project, would it make sense to add an additional option to not use a target project for if someone does not want partial translations filled in?

ddaspit (approved):
We can merge it the way it is. Let's see how it is used and make tweaks if necessary.

> Reviewed 1 of 2 files at r2, 2 of 2 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/233#-:-Nkb4kD_5UUJl5h76Q-K:bihdcuf)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @isaac091)

PR #246, opened 2023-12-01
fix clear cache script

mshannon-sil:
The clear cache script is working now. I checked to make sure the corresponding folder contents were deleted on the AQuA server when the script is run.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/246)

### DIFF
--- a/scripts/clear_clearml_cache.py
+++ b/scripts/clear_clearml_cache.py
@@ -1,29 +1,28 @@
 import glob
 import os
+import shutil
+from pathlib import Path
 
-from clearml import Task
+from silnlp.nmt.clearml_connection import SILClearML
 
-# Warning!  This does not work - it needs to be further debugged.
-
-task = Task.init(
-    project_name="clear_cache",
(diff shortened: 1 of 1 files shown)


mshannon-sil:
Updated the script since the poetry install was happening before the cache got cleared. Since the problem was that it was stalling during the poetry install, that meant it didn't get around to clearing the cache. I changed it so that the cache gets cleared as part of the docker setup bash script, before the poetry install. Now the script clears the caches and then reinstalls the pip packages into the cache.

johnml1135:
Looks great to me!  It should be straightforward enough to run if needed in the future.

johnml1135 (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/246#-:-Nk_gsrW4nTILGtnlItl:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @ddaspit)

PR #255, opened 2023-12-04
Fix inconsistencies between function signatures and calls

isaac091:
I fixed an overcorrection from my first patch that caused a new bug in the translate script and made an adjustment in the test script. Both were issues stemming from mismatches between function signature types and the types of what was actually being passed to the functions.


ddaspit (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/255#-:-Nkq33J874cV60tP3LPu:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @isaac091)

PR #283, opened 2024-01-02
make bettertransformer optional, upgrade to latest version of transformers

mshannon-sil:
BetterTransformer is now configurable rather than always active. The reason for the change is that BetterTransformer doesn't support attention masks for padding, which we use in our experiments during batched training. There is now a `bettertransformer` boolean option under the `train` and `infer` sections of the config.yml for experiments. The default value is set to false.

I also upgraded the transformers package to the most recent version, 4.36.2, since as of version 4.36.0, Scaled Dot Product Attention (SDPA) is now part of transformers directly without needing to convert to BetterTransformer.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/283)

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -953,13 +953,13 @@ files = [
 
 [[package]]
 name = "fsspec"
-version = "2022.10.0"
+version = "2023.12.2"
 description = "File-system specification"
 optional = false
-python-versions = ">=3.7"
+python-versions = ">=3.8"
 files = [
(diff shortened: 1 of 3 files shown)


ddaspit (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/283#-:-NnEv36T4wN2unytjk3I:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @mshannon-sil)

PR #308, opened 2024-01-24
Use torch.compile() when inferencing

isaac091:
Using torch.compile() with the default backend `inductor` causes inferencing to run slightly faster (average of 3.5% improvement across 5 tests) with no impact on memory usage or BLEU score. I also ran tests with other backends (`openxla_eval`, `onnxrt`, `cudagraphs`, and `tvm`), but they all slowed down inferencing.

Compiling after the model is in the pipeline is equivalent to compiling directly after the model is created, but it prevents an irrelevant warning message from appearing.

### DIFF
--- a/silnlp/nmt/hugging_face_config.py
+++ b/silnlp/nmt/hugging_face_config.py
@@ -849,6 +849,7 @@ class HuggingFaceNMTModel(NMTModel):
             tgt_lang=self._config.test_trg_lang,
             device=0,
         )
+        pipeline.model = torch.compile(pipeline.model)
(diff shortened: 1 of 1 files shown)


ddaspit (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/308#-:-Np01pl52cGYbUjm3dWR:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @isaac091)

PR #319, opened 2024-02-06
Madlad Implementation

isaac091:
Obviously I have more testing to do but I think all of the implementation is worked out.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/319)

### DIFF
--- a/silnlp/nmt/hugging_face_config.py
+++ b/silnlp/nmt/hugging_face_config.py
@@ -203,13 +203,13 @@ def delete_tokenizer(checkpoint_path: Path) -> None:
 def add_lang_code_to_tokenizer(tokenizer: PreTrainedTokenizer, lang_code: str) -> None:
     tokenizer.add_special_tokens({"additional_special_tokens": tokenizer.additional_special_tokens + [lang_code]})
     lang_id = tokenizer.convert_tokens_to_ids(lang_code)
-    if isinstance(tokenizer, (NllbTokenizer, MBart50Tokenizer, MBartTokenizer)):
+    if not isinstance(tokenizer, (T5Tokenizer, T5TokenizerFast)):
         tokenizer.lang_code_to_id[lang_code] = lang_id
+    if isinstance(tokenizer, (NllbTokenizer, MBart50Tokenizer, MBartTokenizer)):
         tokenizer.id_to_lang_code[lang_id] = lang_code
         tokenizer.fairseq_tokens_to_ids[lang_code] = lang_id
         tokenizer.fairseq_ids_to_tokens[lang_id] = lang_code
     elif isinstance(tokenizer, M2M100Tokenizer):
-        tokenizer.lang_code_to_id[lang_code] = lang_id
         tokenizer.lang_code_to_token[lang_code] = lang_code
         tokenizer.lang_token_to_id[lang_code] = lang_id
         tokenizer.id_to_lang_token[lang_id] = lang_code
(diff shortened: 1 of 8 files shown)


silnlp/nmt/hugging_face_config.py
```diff
@@ -290,6 +302,10 @@ def __init__(self, exp_dir: Path, config: dict) -> None:
 
         super().__init__(exp_dir, config)
 
+        if self.model.startswith("google/madlad400"):
```
isaac091, line 305: Should I generalize this or leave it as is for now?

silnlp/nmt/hugging_face_config.py
```diff
@@ -290,6 +302,10 @@ def __init__(self, exp_dir: Path, config: dict) -> None:
...
+            self.train["max_source_length"] = 256
+            self.train["max_target_length"] = 256
```
isaac091, line 307: Should I generalize this or leave it as is for now?

ddaspit (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/319)* status: 0 of 1 files reviewed, 4 unresolved discussions (waiting on @isaac091)

___
*[`silnlp/nmt/hugging_face_config.py` line 307 at r1](https://reviewable.io/reviews/sillsdev/silnlp/319#gh-1480564277:-Nq4Igx3ElCJlWxvc0ge:b63p3ae) ([raw file](https://github.com/sillsdev/silnlp/blob/9e1a181a937e97156d6e1d1f0e9b4fc46c21a68f/silnlp/nmt/hugging_face_config.py#L307)):*
<details><summary><i>Previously, isaac091 (Isaac Schifferer) wrote…</i></summary><blockquote>

Should I generalize this or leave it as is for now?
</blockquote></details>

NLLB has `max_length` set in the model config and the generation config. It might be easier to create a function that sets these values for a model after it is created.
___
*[`silnlp/nmt/hugging_face_config.py` line 495 at r1](https://reviewable.io/reviews/sillsdev/silnlp/319#-Nq4J0gf7o24Z4-leZCZ:-Nq4J0gf7o24Z4-leZC_:bzivfv) ([raw file](https://github.com/sillsdev/silnlp/blob/9e1a181a937e97156d6e1d1f0e9b4fc46c21a68f/silnlp/nmt/hugging_face_config.py#L495)):*
> ```Python
>             for iso in self.src_isos | self.trg_isos:
>                 lang_code = lang_codes.get(iso, iso)
>                 if ((not isinstance(self._tokenizer, (T5Tokenizer, T5TokenizerFast)) and lang_code not in self._tokenizer.lang_code_to_id) 
> ```

This `if` statement is difficult to understand. It might be clearer if you split it into separate `if` statements.

```
if isinstance(self._tokenizer, (T5Tokenizer, T5TokenizerFast)):
    if lang_code not in self._tokenizer.all_special_tokens:
        add_lang_code_to_tokenizer(self._tokenizer, lang_code)
        updated = True
elif lang_code not in self._tokenizer.lang_code_to_id:
    add_lang_code_to_tokenizer(self._tokenizer, lang_code)
    updated = True
```
___
*[`silnlp/nmt/hugging_face_config.py` line 746 at r1](https://reviewable.io/reviews/sillsdev/silnlp/319#-Nq4HD7R7KTgY6PmQjYZ:-Nq4HD7R7KTgY6PmQjY_:bun547q) ([raw file](https://github.com/sillsdev/silnlp/blob/9e1a181a937e97156d6e1d1f0e9b4fc46c21a68f/silnlp/nmt/hugging_face_config.py#L746)):*
> ```Python
>             # as the first generated token.
>             forced_bos_token_id = tokenizer.convert_tokens_to_ids(self._config.val_trg_lang)
>             if self._config.model.startswith("google/madlad400"):
> ```

I think the main issue here is that we are always setting `forced_bos_token_id` when we should only be setting it for mBART and M2M-100. The `decoder_start_token_id` should already be set to the pad token.
___
*[`silnlp/nmt/hugging_face_config.py` line 933 at r1](https://reviewable.io/reviews/sillsdev/silnlp/319#-Nq4HkLr31tGqJwzVI9W:-Nq4HkLs0QGoSoFIlNnZ:bl4sreq) ([raw file](https://github.com/sillsdev/silnlp/blob/9e1a181a937e97156d6e1d1f0e9b4fc46c21a68f/silnlp/nmt/hugging_face_config.py#L933)):*
> ```Python
>         else:
>             model_name = self._config.model
>         dtype = torch.bfloat16 if self._is_t5 else torch.float16
> ```

You should create a method that creates the model and use it for the translate methods.

ddaspit (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/319#-:-NqhKkUS2AbCrVk3g4wJ:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @isaac091)

PR #325, opened 2024-02-17
Experiment translate option

isaac091:
Adds --translate option for experiments. Users can define a list translate configs in the config file using the same arguments as the translate script.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/325)

### DIFF
--- a/silnlp/nmt/config.py
+++ b/silnlp/nmt/config.py
@@ -442,10 +442,6 @@ class Config(ABC):
         return self.root["eval"]
 
     @property
-    def translate(self) -> list:
-        return self.root["translate"]
-
-    @property
     def mirror(self) -> bool:
         return self.data["mirror"]
 
--- a/silnlp/nmt/config.py
+++ b/silnlp/nmt/config.py
@@ -442,6 +442,10 @@ class Config(ABC):
         return self.root["eval"]
 
     @property
+    def translate(self) -> list:
+        return self.root["translate"]
+
+    @property
     def mirror(self) -> bool:
         return self.data["mirror"]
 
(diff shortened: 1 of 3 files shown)


ddaspit (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/325)* status: all files reviewed, 1 unresolved discussion (waiting on @isaac091)

___
*[`silnlp/nmt/experiment.py` line 87 at r1](https://reviewable.io/reviews/sillsdev/silnlp/325#-Nr0jyC7CoRqd7yCp6bC:-Nr0jyC8AlPcFBCdJWR_:b11p6mj) ([raw file](https://github.com/sillsdev/silnlp/blob/c0c7b7619a517c3047e6050add18fa59d19db458/silnlp/nmt/experiment.py#L87)):*
> ```Python
> 
>     def translate(self):
>         translate_configs = self.config.translate
> ```

What about the possibility of having the experiment script accept the same CLI arguments as the translate script? I don't love specifying the translate arguments in the experiment config.

mmartin9684-sil:
It's not unusual to need to use the same model to create several draft translations - e.g., from English Bible A to the vernacular, and from English Bible B to the vernacular.

If we want to go with command line arguments for the translate step, would it be possible to support this?

isaac091:
I think having a separate `translate_config.yml` like we talked about this morning makes the most sense. This way it's completely separated from the training and we're still able to have multiple translation requests in one go.

ddaspit (approved):
I like it.

> Reviewed 3 of 3 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/325#-:-Nr6OF316w11FYLO0y2X:b8eau52)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @isaac091)

PR #326, opened 2024-02-20
Slightly more robust syncing

davidbaines:
Added a check so that the investigations could be processed without a /.
At the moment this shouldn't be required, but if we were to ever separate out the Bleu scores from all the extra details it would be required.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/326)

### DIFF
--- a/clowder/investigation.py
+++ b/clowder/investigation.py
@@ -405,7 +405,7 @@ class Investigation:
                 out_df = out_df[["score"]]                
                 out_df = out_df.transpose()
                 out_df.columns = pd.Index(["BLEU"])
-                out_df["BLEU"] = out_df["BLEU"].apply(lambda x: x.split("/")[0])
+                out_df["BLEU"] = out_df["BLEU"].apply(lambda x: str(x).split("/")[0] if '/' in str(x) else str(x))
                 out_df["BLEU"] = out_df["BLEU"].apply(pd.to_numeric)
                 out_df[["spBLEU", "CHRF3", "WER", "TER","BLEU-details"]] = None
                 out_df["NumberOfSteps"] = num_steps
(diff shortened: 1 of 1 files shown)


Enkidu93 (changes requested):
Thank you for doing this! I apologize that the timezone difference makes being responsive tricky.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/326#-:-Nr5PnCl7WMPMJ_2t4o2:bnr0kgs)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @davidbaines)

___
*[`clowder/investigation.py` line 408 at r1](https://reviewable.io/reviews/sillsdev/silnlp/326#-Nr5PJpV2RrviUp-db2t:-Nr5PJpV2RrviUp-db2u:btlmke7) ([raw file](https://github.com/sillsdev/silnlp/blob/8f714dc05b86577f6689acd1b707b5a2d63a6a92/clowder/investigation.py#L408)):*
> ```Python
>                 out_df = out_df.transpose()
>                 out_df.columns = pd.Index(["BLEU"])
>                 out_df["BLEU"] = out_df["BLEU"].apply(lambda x: str(x).split("/")[0] if '/' in str(x) else str(x))
> ```

Since the behavior of split is such that if the delimiter is not in the string, it will just return a single-element list containing the entire string, you shouldn't need to check for the presence of '/'. So, `lambda x: str(x).split('/')[0]` should work for values like `3`, `'3'`, and `'3/6/7'`.

davidbaines:
Thanks Eli. I made that change and pushed them.

Enkidu93 (approved):
:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/326#-:-NrB0s7vC8Ze6CFgPBJp:bg7igkv)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @davidbaines)

Enkidu93 (commented):
> Dismissed @Enkidu93 from a discussion.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/326)* status: 0 of 1 files reviewed, all discussions resolved

___
*[`clowder/investigation.py` line 408 at r1](https://reviewable.io/reviews/sillsdev/silnlp/326#-Nr5PJpV2RrviUp-db2t:-NrB1E-TCTwxIYNJRiUC:b-896fix) ([raw file](https://github.com/sillsdev/silnlp/blob/8f714dc05b86577f6689acd1b707b5a2d63a6a92/clowder/investigation.py#L408)):*

Since the behavior of split is such that if the delimiter is not in the string, it will just return a single-element list containing the entire string, you shouldn't need to check for the presence of '/'. So, `lambda x: str(x).split('/')[0]` should work for values like `3`, `'3'`, and `'3/6/7'`.
</blockquote></details>

Done.

PR #331, opened 2024-02-23
Fix bug when adding multiple language codes

isaac091:
`add_special_tokens` will not add a token if that token is already in the special tokens, and since the default behavior is to replace the whole special tokens list with the new list, passing the current list of special tokens effectively cancels them out, leaving only the most recent token. This was fine before because only one new language code is typically added for NLLB experiments, but for MADLAD, language codes are not special tokens by default and so adding both the source and target language causes the issue.

I'm actually going to stop adding the source language token for MADLAD because it isn't used, but the fix is still needed for any time >1 new language code is added.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/331)

### DIFF
--- a/silnlp/nmt/hugging_face_config.py
+++ b/silnlp/nmt/hugging_face_config.py
@@ -201,7 +201,7 @@ def delete_tokenizer(checkpoint_path: Path) -> None:
 
 
 def add_lang_code_to_tokenizer(tokenizer: PreTrainedTokenizer, lang_code: str) -> None:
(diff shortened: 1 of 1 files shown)


ddaspit (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/331#-:-NrMrdGy9Yx7XjlT1IOn:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @isaac091)

Issue #340, opened 2024-03-06
Handling "low surrogate" characters in KBTFRA SFM files


mmartin9684-sil:
Extracting the KBTFRA project or running the translate command on SFM files in the KBTFRA project results in 'invalid continuation byte' error messages (see below).

The issue seems to be traceable to the use of 'low surrogate' characters in the project SFM files, as shown in an extract from a _Wildebeest_ character checking report below.  Running the _Wildebeest_ character normalization script over all of the SFM files in the project results in SFM files that can be extracted and translated.  (See S: > Paratext > projects > KBTFRA_norm.)

Stack trace - extract_corpora command:
```
(silnlp-py3.8) PS C:\Users\mm9q\PycharmProjects\silnlp> poetry run python -m silnlp.common.extract_corpora KBTFRA
2024-03-06 08:50:37,595 - silnlp.common.environment - INFO - Using workspace: S:/ as per environment variable SIL_NLP_DATA_PATH.
2024-03-06 08:50:37,825 - silnlp.common.extract_corpora - INFO - Extracting KBTFRA...
Traceback (most recent call last):
  File "C:\Users\mm9q\AppData\Local\Programs\Python\Python38\lib\runpy.py", line 194, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "C:\Users\mm9q\AppData\Local\Programs\Python\Python38\lib\runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "C:\Users\mm9q\PycharmProjects\silnlp\silnlp\common\extract_corpora.py", line 105, in <module>
    main()
  File "C:\Users\mm9q\PycharmProjects\silnlp\silnlp\common\extract_corpora.py", line 78, in main
    check_versification(project_dir)
  File "C:\Users\mm9q\PycharmProjects\silnlp\silnlp\common\paratext.py", line 557, in check_versification
    nt_versification, key_nt_verses = detect_NT_versification(project_dir)
  File "C:\Users\mm9q\PycharmProjects\silnlp\silnlp\common\paratext.py", line 510, in detect_NT_versification
    jhn_6 = get_last_verse(project_dir, "JHN", 6)
  File "C:\Users\mm9q\PycharmProjects\silnlp\silnlp\common\paratext.py", line 454, in get_last_verse
    for line in book_file:
  File "C:\Users\mm9q\AppData\Local\Programs\Python\Python38\lib\codecs.py", line 322, in decode
[...]

PR #350, opened 2024-03-22
Fix bug in get_last_verse

isaac091:
Bethany had an issue like #340 as well. I was able to resolve the issue for that project by using the project's encoding, but the same did not work for the KBTFRA project because there was another error of the same type when using the project's encoding. Ignoring the errors should solve the problem because not every character needs to be present to determine the versification.

### DIFF
--- a/silnlp/common/paratext.py
+++ b/silnlp/common/paratext.py
@@ -449,7 +449,7 @@ def get_last_verse(project_dir: str, book: str, chapter: int) -> int:
     last_verse = "0"
     book_path = get_book_path(project_dir, book)
     try:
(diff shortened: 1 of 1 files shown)


ddaspit (approved):
I think this fix is fine for now. We should switch to using the `ParatextProjectSettings` class once it has been ported to Machine.py. This will allow us to use the appropriate encoding.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/350#-:-NtqQxzu0vVj5LP5_qQq:buv42ob)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @isaac091)

PR #363, opened 2024-04-16
Score file format cleanup

isaac091:
Updated score files to have only one line for each data pair. There is a column for each metric whether it's used or not. I also updated the default scorers for the experiment script to include all the chrF scores and none of the error rate ones.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/363)

### DIFF
--- a/clowder/investigation.py
+++ b/clowder/investigation.py
@@ -319,7 +319,7 @@ class Investigation:
                                     steps_element = name_elements[1][:-4]
                                     if steps_element.isdigit():
                                         num_steps = int(steps_element)
-                            df = self._process_scores_csv(df, num_steps)
+                            df.insert(len(df.columns), "num_steps", np.repeat(num_steps, len(df.index)))
                         df.insert(0, NAME_ATTRIBUTE, np.repeat(row[NAME_ATTRIBUTE], len(df.index)))
                         if name not in results:
                             results[name] = pd.DataFrame()
@@ -417,40 +417,6 @@ class Investigation:
                 task: Optional[Task] = Task.get_task(task_id=clearml_id)
                 tasks[experiment_name] = task
         return tasks
-
-    def _process_scores_csv(self, df: pd.DataFrame, num_steps: str) -> pd.DataFrame:
-        groups = df.groupby(["src_iso", "trg_iso"]).groups
-        dfs = []
-        for (src_iso, trg_iso), _ in groups.items():
-            out_df = df.loc[(df["src_iso"] == src_iso) & (df["trg_iso"] == trg_iso), :]
-            if src_iso == "ALL" and trg_iso == "ALL":
-                out_df = out_df[["score"]]
-                out_df = out_df.transpose()
-                out_df.columns = pd.Index(["BLEU"])
-                out_df["BLEU"] = out_df["BLEU"].apply(lambda x: str(x).split("/")[0])
-                out_df["BLEU"] = out_df["BLEU"].apply(pd.to_numeric)
-                out_df[["spBLEU", "CHRF3", "WER", "TER", "BLEU-details"]] = None
-                out_df["NumberOfSteps"] = num_steps
-                out_df = out_df[["BLEU", "spBLEU", "CHRF3", "WER", "TER", "NumberOfSteps", "BLEU-details"]]
-            else:
-                out_df = out_df[["score", "scorer"]]
-                column_names = out_df[["scorer"]].values.flatten()
-                out_df = out_df[["score"]]
-                out_df = out_df.transpose()
-                out_df.columns = pd.Index(column_names)
-                out_df["BLEU-details"] = out_df["BLEU"]
-                out_df["BLEU"] = out_df["BLEU"].apply(lambda x: str(x).split("/")[0])
-                out_df[["BLEU", "spBLEU", "CHRF3", "WER", "TER"]] = out_df[
-                    ["BLEU", "spBLEU", "CHRF3", "WER", "TER"]
-                ].apply(
-                    pd.to_numeric, axis=0
-                )  # TODO more robust (ignore for mvp)
-                out_df["NumberOfSteps"] = num_steps
-                out_df = out_df[["BLEU", "spBLEU", "CHRF3", "WER", "TER", "NumberOfSteps", "BLEU-details"]]
-            out_df.insert(0, "src_iso", src_iso)
-            out_df.insert(0, "trg_iso", trg_iso)
-            dfs.append(out_df)
-        return pd.concat(dfs, join="outer", ignore_index=True)
 
     def _min_and_max_per_col(self, df: pd.DataFrame):
         df = df.select_dtypes(include="number")
--- a/silnlp/nmt/experiment.py
+++ b/silnlp/nmt/experiment.py
@@ -164,7 +164,7 @@ def main() -> None:
         nargs="*",
         metavar="scorer",
         choices=_SUPPORTED_SCORERS,
-        default=["bleu", "sentencebleu", "chrf3", "chrf3++", "wer", "ter", "spbleu"],
+        default=["bleu", "sentencebleu", "chrf3", "chrf3+", "chrf3++", "spbleu"],
         help=f"List of scorers - {_SUPPORTED_SCORERS}",
     )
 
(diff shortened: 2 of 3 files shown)


Enkidu93:
*[`silnlp/nmt/test.py` line 51 at r1](https://reviewable.io/reviews/sillsdev/silnlp/363#-Nvggsx7CZ7G4Rjuvc2F:-NvgnwAZ9lmGv2wrlnoc:b76hbla) ([raw file](https://github.com/sillsdev/silnlp/blob/b5b55b6f0b0e60246d73f9815e8f5f5cdb933b26/silnlp/nmt/test.py#L51)):*

What happens when we add a new scorer? Do we add another column at the end? Will the current tools that parse this file be able to handle this?
</blockquote></details>

In regards to Clowder, this will be a breaking change (but ultimately, a very nice simplification). Would it be best for me to make the changes to Clowder in this same branch as part of the same PR? (Out of curiosity, what other tools parse this file?)

ddaspit (commented):
I don't have a problem if you want to make the columns dynamic. Clowder will need to handle that properly. If tools are written to handle dynamic scorer columns, then they will be able to handle it when a new scorer is added. It might be helpful to indicate in the column header which columns are for scorers.

David's tools are the tools I was thinking of. I'm not aware of any other tools.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/363#-:-NvhmXE8Ahyr6GJ_dDU7:blokvjm)* status: all files reviewed, 1 unresolved discussion (waiting on @Enkidu93)

___
*[`silnlp/nmt/test.py` line 51 at r1](https://reviewable.io/reviews/sillsdev/silnlp/363#-Nvggsx7CZ7G4Rjuvc2F:-Nvhm6bP9w2MSt04aJO5:b-20udk1) ([raw file](https://github.com/sillsdev/silnlp/blob/b5b55b6f0b0e60246d73f9815e8f5f5cdb933b26/silnlp/nmt/test.py#L51)):*

In regards to Clowder, this will be a breaking change (but ultimately, a very nice simplification). Would it be best for me to make the changes to Clowder in this same branch as part of the same PR? (Out of curiosity, what other tools parse this file?)
</blockquote></details>

You don't need to make the Clowder changes in this PR if you don't want to. I will leave it to you to decide.

isaac091:
One small issue with having dynamic columns is that because the `write` function is a method of the PairScore class and therefore can't see what scorers other PairScores are using, the BLEUDetails column of the final summary line for experiments with multiple data pairs would be misaligned. My current solution to this is a bit hacky, I'm basically just copying some other scores into the summary PairScore and then in `write`, if it's a summary, I add a column but not the score for each extra scorer.

Would it be better to actually just calculate the other summary scores? Would this information be useful or would it just be clutter? Is there a specific reason why only a summary BLEU is calculated?

ddaspit (commented):
Is there a reason why you changed the BLEU score output, so it is split into two columns? If you merge it back to a single score, then this is no longer a problem. For other BLEU-based scorers, such as spBLEU, it might be valuable to include all of the details.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/363#-:-NvrDV9B9mLikSQlMLxx:b1uqnjs)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @Enkidu93)

isaac091:
Good catch, I was working under the assumption that it was like that originally, I forgot that that was part of Eli and I's original revision. I blame going to the conference in the middle of this haha. I don't want to speak for Eli, but I think we did it to make it easier to quickly see all the other scores, since the extra numbers aren't used as much. I'll change it back, since that's really just a preference thing. I'll also add the rest of the spBLEU numbers.

Enkidu93:
I have no issue with dynamic columns if the format is fixed. But I'm confused - is there still a BLEUdetails column? (I see references below in the code). I'm happy to be outvoted, but I really don't like the BLEU score as a string of values separated by slashes - it adds complexity if you want to use BLEU as a numeric value and there's no header information to inform you about the meanings of those different values. I don't know if this is a possibility, but if we can't have BLEU and BLEUdetails, could we just have separate columns for each of those BLEU detailed values with a meaningful column name? If not, I can do that processing in clowder, but it does seem like it would be an improvement in general.

isaac091:
I haven't made the commit yet that would get rid of the BLEUDetails column. That's a good point about being able to use the column as a number without extra processing, and it would be easy to make separate columns for the BLEU score. My only concern is that having a bunch of extra columns would make it more difficult to quickly look at all the scores and know what you're looking at. But as long as you think that won't be too much of an issue, that's fine with me.

If we do decide to break the numbers out into multiple columns, would we want to do the same with any additional BLEU-based scores? My initial thought is no because we don't want to explode the number of columns, but having a mix of formats for the same type of score could be confusing.

ddaspit (commented):
I think we should ask Michael, Bethany, and David if they want the BLEU details or if they would be happy with the single score. If they want the BLEU details, then we can split it out into separate columns. Also, if you need to refactor it so that the `PairScore` class is no longer responsible for writing the rows, that is fine.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/363#-:-NvrsTNZ0sklXKl1Qr9_:bj7d355)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @Enkidu93)

ddaspit (commented):
It might be good to prefix the BLEU details with `BLEU_`.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/363#-:-NwBIYZTE8V8CHo-dg8i:b-y8d34)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @Enkidu93)

ddaspit (commented):
@Enkidu93, would it be helpful if the score columns were indicated as such in the header? This would make it easier to separate the potentially dynamic score columns from the fixed columns.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/363#-:-NwBIokA7X4Stj18gyuk:b-5qbwf5)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @Enkidu93)

Enkidu93:
> @Enkidu93, would it be helpful if the score columns were indicated as such in the header? This would make it easier to separate the potentially dynamic score columns from the fixed columns.
> 
> > _[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/363#-:-NwBIokA7X4Stj18gyuk:b-5qbwf5)_ status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @Enkidu93)

I'm not sure I understand what you mean. Could you give an example output? Do you mean a multiheader or adding something to the column name or what?

ddaspit (commented):
I just mean adding something to the column name, e.g. "BLEU-score".

> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/363#-:-NwBYUMBB7Sgem-Bi609:b-whdmmk)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @Enkidu93)

Enkidu93:
> I just mean adding something to the column name, e.g. "BLEU-score".
> 
> > _[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/363#-:-NwBYUMBB7Sgem-Bi609:b-whdmmk)_ status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @Enkidu93)

Oh, I see - I don't think I have a need for that - at the moment, anyways. 

I've gone ahead and pushed a fix for clowder. Done on my end.

ddaspit (approved):
:lgtm:

> Reviewed 1 of 1 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/363#-:-NwKfPQf3OI1tGxvsCuT:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @Enkidu93)

ddaspit (commented):
> Reviewed 1 of 1 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/363)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @Enkidu93)

PR #372, opened 2024-04-26
Lora weight tying update

isaac091:
I was having trouble tying the embeddings to the lm_head in LoRA models where the embeddings were actually targeted layers, not just unfrozen, because the embeddings become a specific `lora_embedding` module, whereas the lm_head is treated normally, and the two aren't quite compatible. I didn't spend a ton of time on it because I know we mostly want to fully train the embeddings, so if we ever want to target the embeddings and keep them tied together, there is some more work to be done.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/372)

### DIFF
--- a/silnlp/nmt/hugging_face_config.py
+++ b/silnlp/nmt/hugging_face_config.py
@@ -1152,6 +1152,18 @@ class HuggingFaceNMTModel(NMTModel):
             elif self._config.model_prefix == "google/madlad400":
                 embedding = model.base_model.model.encoder.embed_tokens.modules_to_save.default.weight
                 model.base_model.model.decoder.embed_tokens.modules_to_save.default.weight = embedding
+        elif "embed_tokens" in target_modules:
+            if self._config.model_prefix == "facebook/nllb-200":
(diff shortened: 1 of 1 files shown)


ddaspit (changes requested):
Good job. I think it would be good to capture the issue about targeting embeddings and the lm_head when they are tied in a TODO comment.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/372#-:-NwQMVvV0_7LyV43Sn3L:bhjzzom)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @isaac091)

isaac091:
I ended up adding the code for tying targeted embeddings because I already had it and MADLAD doesn't actually need the fix to work like it's supposed to.

ddaspit (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/372#-:-NwQ_J6P7x9oO5Fitlfh:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @isaac091)

PR #383, opened 2024-05-22
Save LoRA models as full merged models rather than adapters

isaac091:
The only changed code that should be run before the first `_save` call is the `create_tied_embedding_weights` function, which got moved outside of the `HuggingFaceNMTModel` class, and when that function is called, now after the `lora_config` is loaded, and I did verify that the model was set up the same way it was before, i.e. that all the right weights were tied in the right ways. Previously, the losses, eval BLEU scores, and test scores were identical between reruns of the same experiment, so I'm not sure what could be causing this additional non-determinism.
<br>
A couple other pieces of information I didn't mention earlier that may or may not be useful:
* Most of the (identically configured) experiments I ran with these changes had the same best/last checkpoints (7k/11k), but a couple were different (one was 6k/10k, another was 8k/9k).
* While all of the test scores have been within a point of the scores of the original LoRA experiment, they have all been lower. It's possible that this is just a coincidence, but I did run the same experiment 5 times.
* When I've experimented with how LoRA models have been saved in the past, the test scores were always either all 0/very low, or the same scores as the original LoRA model.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/383)

### DIFF
--- a/silnlp/nmt/hugging_face_config.py
+++ b/silnlp/nmt/hugging_face_config.py
@@ -3,6 +3,7 @@ import json
 import logging
 import os
 import re
+import shutil
 from contextlib import ExitStack
 from copy import deepcopy
 from enum import Enum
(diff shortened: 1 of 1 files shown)


ddaspit (commented):
I don't see anything that could affect the training other than that you check to see if `embed_tokens` is set in `modules_to_save` or `target_modules` before updating the tied embeddings. What happens if you train a LoRA model with the current master branch? Are you seeing the same drop in score? Does it only occur on this branch?

> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/383#-:-NybKyk112x4Z6tppJ6N:byu3aqf)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @isaac091)

isaac091:
Update: By waiting to merge the LoRA adapter into the model until after training, I was able to get the reduced model size without affecting training. I'm still not sure what part of merging in the save function (on a separate model) was messing with the training. 

Using the same approach for MADLAD models results in the full 11.7 GB model size, but the adapters are only 3.2 GB, so they aren't merged. I'm not sure what the difference is between the two models that causes NLLB to have inflated adapter sizes. I can see in the code that PEFT saves a copy for each instance of a shared module, while HF only saves one copy, so I assumed this was causing the difference in adapter sizes, but that must not be the only factor. I assume there is some other difference in the models' implementations, but I couldn't find anything.

ddaspit (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/383#-:-O2aESvcD6GuU3-meBR0:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @isaac091)

Issue #387, opened 2024-05-28
Fix warning for gradient checkpointing by specifying value for use_reentrant


mshannon-sil:
When running NMT training in SILNLP, the following warning is appearing:
`torch.utils.checkpoint: please pass in use_reentrant=True or use_reentrant=False explicitly. The default value of use_reentrant will be updated to be False in the future. To maintain current behavior, pass use_reentrant=True. It is recommended that you use use_reentrant=False. Refer to docs for more details on the differences between the two variants.`

We'll need to to conduct some experiments to determine whether there is any accuracy or performance difference between the two options. And based on the results, we'll need to set the default in the config.

mshannon-sil:
From the five experiments I conducted with early stopping, setting use_reentrant to True slightly increased the BLEU score by 0.94 on average. And the training time results were inconsistent, with True training faster three times and False training faster two times. So for now, I'll plan to set it to True by default, especially since this is the setting we've already been using.

The only potential concern I've seen discussed is that it can cause issues when freezing layers or using LoRA. This is due to the following requirement for the True version: `At least one input and output must have requires_grad=True for the reentrant variant. If this condition is unmet, the checkpointed part of the model will not have gradients. The non-reentrant version does not have this requirement.` If we do run into issues in those situations, we can set use_reentrant to False in the experiment's config.yml. I'll add documentation for this to the wiki.

PR #403, opened 2024-06-05
set use_reentrant to true

mshannon-sil:
Per the discussion in issue #387, I'm setting use_reentrant to True by default to fix the warning, since it seems to perform slightly better than the False setting and is also the setting that we have been using, just not explicitly. If we run into any issues with freezing layers, the default can be overridden in the experiment's config.yml. When this pull request is approved, I'll go ahead and update the wiki with that info for the config.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/403)

### DIFF
--- a/silnlp/nmt/hugging_face_config.py
+++ b/silnlp/nmt/hugging_face_config.py
@@ -102,6 +102,7 @@ TRAINING_ARGS_CONFIG_MAPPING = {
     "train": {
         "gradient_accumulation_steps",
         "gradient_checkpointing",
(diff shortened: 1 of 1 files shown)


ddaspit (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/403#-:-NzdIFOH2s8J-8945eyv:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @mshannon-sil)

PR #404, opened 2024-06-05
Utilities

davidbaines:
count_verses.py first reads previous results and calculates verse counts for new scriptures. 
It has default arguments so can typically be run without specifying which files to process.
Results are saved to MT/experiments/verses/verse-counts.csv and to an Excel file.

Books in the Deuterocanon are explicitly ignored.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/404)


davidbaines:
I've added a few utilities. None of them should impact any of the other code in in the repo.

**check_books** takes a project and a list of book ID's and tests to see whether they can be parsed by SILNLP. Without a list of books it will check all the books in the project. Typical use is without the --verbose, and without a list of books, to find which books can't be parsed. Then run with one book and the --verbose option to find the error that prevented parsing. Ctrl+click the file to jump to the line that couldn't parse.

**combine_scores** takes an experiment series folder as an argument and combines all the scores into one scores.csv file and also writes scores.xlsx.  The results are written to the series folder. It copes with the two different styles of scores files, splitting them into two sheets in Excel.

**count_verses** Uses S:\MT\scripture as the default source folder and S:\MT\experiments\verses\verses.csv as its default output file. Typically no arguments are required.  It reads the current verses.csv file, calculates the verse counts for new files and updates the output file.

ddaspit (commented):
> Reviewed 5 of 5 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/404)* status: all files reviewed, 1 unresolved discussion (waiting on @bhartmoore, @davidbaines, and @Enkidu93)

___
*[`silnlp/common/extract_corpora.py` line 57 at r3](https://reviewable.io/reviews/sillsdev/silnlp/404#-Nzo1BtoCy2C-Rm4M70W:-Nzo1BtoCy2C-Rm4M70X:b-2pmws1) ([raw file](https://github.com/sillsdev/silnlp/blob/1857d6fff7c12e288caa37e87cfd3c199de16a9c/silnlp/common/extract_corpora.py#L57)):*
> ```Python
>     projects_found = []
>     for project_pattern in project_patterns:
>         matching_projects = project_path.glob(f"*{project_pattern}*")
> ```

Does this work with the S3 bucket?

davidbaines:
Yes it works on the S3 bucket, on my machine at least.
I'm surprised to see that change here. I thought I had only made changes to extract_corpora on my local machine and hadn't pushed them yet.

ddaspit (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/404)* status: 3 of 7 files reviewed, 2 unresolved discussions (waiting on @bhartmoore, @davidbaines, and @Enkidu93)

___
*[`silnlp/common/extract_corpora.py` line 121 at r5](https://reviewable.io/reviews/sillsdev/silnlp/404#-O-31Z5D2ui-NlJQ7XQQ:-O-31Z5D2ui-NlJQ7XQR:b-ome2r7) ([raw file](https://github.com/sillsdev/silnlp/blob/12c31b50ac28e1e32a797258f5deb348beb51c00/silnlp/common/extract_corpora.py#L121)):*
> ```Python
>         f"Waiting {seconds} seconds for the files to arrive in {SIL_NLP_ENV.mt_scripture_dir} before counting the verses."
>     )
>     for second in tqdm(range(seconds)):
> ```

Why do we need to wait 30 seconds?

davidbaines:
I tried not waiting and the files that were being written to the S: drive
weren't there, so it just reports that there are no new files.
I tried waiting 15 seconds and that still wasn't long enough. I settled for
30 seconds. It's admittedly very clunky and I thought about removing that
code altogether. Though I think the convenience of not having to put in
another command is useful.

davidbaines:
The wait wasn't necessary, it was looking in the wrong folder.  Corrected that and removed the wait.

davidbaines:
I'll split this into smaller chunks and they'll be easier to review that way.

PR #451, opened 2024-07-11
Add ability to specify projects as back translations

### LINKED ISSUE #377: Back Translation language code management
We need to be able to have a mixed source corpus pair containing a published Bible and a back translation where there are separate test sets for each source and there are no verses from the published Bible's test set in the back translation's training set. This can be done by using a different language code for each source (while still mapping each code to the same NLLB language code), but there should be a more official way to set up an experiment in this way.

isaac091:
Adds an optional field called `back_translations` to corpus pairs whose value is a list of projects in the same format as `src` or `trg`. Projects in this list have "_BT" appended to their language code so that their test set is evaluated separately. The new language code is still mapped to the correct NLLB language code.

Closes #377.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/451)

### DIFF
--- a/silnlp/nmt/config.py
+++ b/silnlp/nmt/config.py
@@ -94,6 +94,7 @@ class DataFile:
 class CorpusPair:
     src_files: List[DataFile]
     trg_files: List[DataFile]
+    back_translations: List[str]
     type: DataFileType
     src_noise: List[NoiseMethod]
     augmentations: List[AugmentMethod]
@@ -176,6 +177,9 @@ def parse_corpus_pairs(corpus_pairs: List[dict]) -> List[CorpusPair]:
         if isinstance(trg, str):
             trg = trg.split(",")
         trg_files = [DataFile(get_mt_corpus_path(tp.strip())) for tp in trg]
+        back_translations: Union[str, List[str]] = pair.get("back_translations", [])
+        if isinstance(back_translations, str):
+            back_translations = back_translations.split(",")
         is_scripture = src_files[0].is_scripture
         if not all(df.is_scripture == is_scripture for df in (src_files + trg_files)):
             raise RuntimeError("All corpora in a corpus pair must contain the same type of data.")
@@ -229,6 +233,7 @@ def parse_corpus_pairs(corpus_pairs: List[dict]) -> List[CorpusPair]:
             CorpusPair(
                 src_files,
                 trg_files,
+                back_translations,
                 type,
                 src_noise,
                 augmentations,
@@ -345,6 +350,8 @@ class Config(ABC):
         self.src_projects: Set[str] = set()
         self.trg_projects: Set[str] = set()
         for corpus_pair in self.corpus_pairs:
+            for sf in corpus_pair.src_files:
+                sf.iso += "_BT" if sf.path.stem in corpus_pair.back_translations else ""
             pair_src_isos = {sf.iso for sf in corpus_pair.src_files}
             pair_trg_isos = {tf.iso for tf in corpus_pair.trg_files}
             self.src_isos.update(pair_src_isos)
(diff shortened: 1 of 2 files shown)


Enkidu93 (approved):
:lgtm:

This shouldn't cause any issues if someone is doing a vernacular -> BT draft and includes the BT in the back_translations field, right? If not, even so, I'm just wondering if there's any point in making the field name more specific to the issue that this is meant to address. 

> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/451#-:-O1XGzQyAJDmQtEzW7mE:blrj5lu)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @ddaspit)

isaac091:
No issues with vernacular -> BT because it only looks for BTs in the src field. Name is updated to be more specific.

Enkidu93 (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/451)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @ddaspit)

ddaspit (commented):
This seems a bit hacky to me. I would like to understand the issue a bit better to see if there is a more general solution. We have a mixed source configuration with the main source translation and the back translation. We only want to use the back translation in the training split, but not in the test split. Is that correct? I wonder if we could do something simple like only include the first source project in the test split.

> Reviewed 1 of 2 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/451#-:-O2QDXL2CHg98ckeNjGI:bhsx3jq)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @isaac091)

isaac091:
That's correct. The way I have it set up now, the BT is being added to the test set for the purpose of not having any verses from the test set show up in the BT training set, and the different language code makes it so that they are scored separately.

I looked into how we could avoid adding the BT to the test set altogether, and it looks like it is possible by only checking if the source is a back translation right before adding its test set. This way, test indices can still be defined even if the BT is listed first.

Is there any value in having the BT test set scored (separately)?

ddaspit (commented):
This is much nicer. Good job.

> Reviewed 2 of 2 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/451#-:-O2fmP5oDuaK-cAdGJcN:bp0op7r)* status: all files reviewed, 1 unresolved discussion (waiting on @isaac091)

___
*[`silnlp/nmt/config.py` line 180 at r3](https://reviewable.io/reviews/sillsdev/silnlp/451#-O2fmVH_1P18z90YKtJD:-O2fmVH_1P18z90YKtJE:b-g56a3o) ([raw file](https://github.com/sillsdev/silnlp/blob/36ef2ac2280f3939e8399a00fac58a1f0c2289a2/silnlp/nmt/config.py#L180)):*
> ```Python
>             trg = trg.split(",")
>         trg_files = [DataFile(get_mt_corpus_path(tp.strip())) for tp in trg]
>         back_translations: Union[str, List[str]] = pair.get("mixed_src_BTs", [])
> ```

The config should be more generic. I think we should support data files with properties in the config. It would look something like this:
```
data:
  corpus_pairs:
  - src:
    - name: aaa-BT
      include_test: false
    - aaa-SRC
    trg: bbb-TRG
    type: train, test
```

isaac091:
Makes sense to me. I made the changes and I have a couple of thoughts/questions.
1. Does it make more sense for the processing work to be done in `parse_corpus_pairs` like it is now, or should I move it to a DataFile constructor?
2. Should the `is_scripture` property be converted to a regular field that can be configured? I've noticed that basically all of the data we work with gets put in the `MT/scripture` directory regardless of it being scripture or not, and that's the only qualification for the property to be true. The default could still be "file in scripture dir", but it might be nice to have the option to override.

PR #462, opened 2024-07-22
Handle and sync remote metas

pandanautinspace:
Addresses #427

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/462)


Enkidu93 (commented):
Thank you for doing this!

One general comment: Are you using black as a formatter? We should make sure our formatters are configured similarly so we don't have so many such differences in commits. If you do see unformatted code, it's easier to put that into a separate PR or at least separate commits so it's easier to review - i.e., see the changes to the logic separate from formatting changes.

I'm still hesitant about the strategy here. Doesn't this cause quite a performance hit? It also adds complexity. If it's necessary, it's necessary. Could you explain why this strategy is necessary? I was picturing something more like preferring the remote meta on `sync` and only overwriting the remote meta on functions like `run` within the command itself. (And we could always add a flag to `sync` to force an overwrite if we see that being useful for debugging or something). As long as actions are wrapped in syncs, we shouldn't have the issue with outdated changes overwriting newer ones in the remote meta when there are multiple users, right? If the only changes to the local meta are within functions like run? Also, you don't necessarily need to know if there is an error in fetching investigation info up front - you could just display that on the `InvestigationDetails.py` page, so I don't think that would be an issue. What do you think?

(If it's easier to meet and discuss, let me know!) Thank you for your hard work!

> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/462#-:-O2VhqIW64MLwjr9WnEv:b-ml6ug4)* status: 0 of 4 files reviewed, all discussions resolved

pandanautinspace:
Hey! I don't think the performance hit should be very extreme - we're adding one more field to the YAML file when the initial meta download happens, but the default is not to sync from the remote meta.
The only other place where the remote_sync version of get_investigation happens is in `sync` - so we just check for updates on the remote, where we already will be loading that remote version.

You're right on the error stuff in the onboarding dashboard - but if an investigation doesn't exist, I think it makes sense for it to be clear from the investigation list what is going on, since nothing useful can be done with the investigation from the `InvestigationDetails` page. That way you can just delete the local copy when that occurs, and if there are errors in retrieving the actual details of the investigation they can be shown on the `InvestigationDetails` page.

Enkidu93 (commented):
OK, that makes sense. My concern was more about the `investigations` property now actually being a more convoluted operation than just a look up, but I see that it's only used in `status` and `sync` so I guess that's OK. We'll just want to watch out for using it anywhere unnecessarily.

Does this work as expected in the `functions.sync` function? It's a little odd because now the property is updating the investigations from the remote meta but then in the `Investigation.sync` function, the remote meta is getting altered by the local meta. Is there duplicate work being done? Do we need to maybe rethink using a property or maybe use a more descriptive name?

OK, I'm fine with that then as long as it looks alright in the UI.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/462#-:-O2WXvwe85wcYBiYxSk9:b1us1ad)* status: 0 of 4 files reviewed, all discussions resolved

Enkidu93:
:partying_face: Good job!

PR #496, opened 2024-08-23
Find_by_iso needs the languageFamilies.json file. 

davidbaines:
The languageFamilies.json data is necessary for the find_by_iso code. find_by_iso occasionally helps in finding extracts that align well with a given target extract which leads to a better mixed model.

I'll want to either add code for creating the languageFamilies.json file, or find a way to obtain the data from ethnologue. more directly but this is already useful as is and the data doesn't change very often.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/496)


davidbaines:
I don't know where all the changes and commits came from. I thought I was simply adding one file to the assets folder. 
I'll close this and try again.

Issue #473, opened 2024-08-03
Create initial xri_etl script


rminsil:
# Overview

Create the initial script defined by the parent issue: https://github.com/sillsdev/silnlp/issues/472

The parent issue has most of the details around how the script will work, so I won't repeat that all here.

The scope for this issue is just a basic POC script that produces the `*.all.txt` and `*.(train/val/test).txt` files. More complex things like data transformation will be left for later tickets.

rminsil:
I accidentally closed this.
I intended to put a few PR's under this issue.

A new one is here: https://github.com/sillsdev/silnlp/pull/491

rminsil:
Another PR: https://github.com/sillsdev/silnlp/pull/493

PR #514, opened 2024-09-12
Issue 473 Add optional debug/info logging to extract_xri script

rminsil:
This PR adds optional logging to the script.

Currently the script makes certain assumptions about the data which can make it a bit brittle causing it to crash mid-processing.

Some example assumptions:

- all columns exist with the expected schema
- no embedded newlines (e.g. each row is on one line)
- id's can all be parsed to `int`

Logging will help to quickly identify the row of data breaking it.

Currently the logging is off by default. Enabling with `-log [PATH]` sends the logs to the file specified.

Note this PR is built on top of another unmerged PR #473. I'll rebase this and target master when #473 is merged.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/514)

### DIFF
--- a/silnlp/common/extract_xri.py
+++ b/silnlp/common/extract_xri.py
@@ -32,11 +32,14 @@ The prefix of the filename is built from the cli inputs. For the example above t
 
 Run with --help for more details.
 
+Logging can be enabled by adding an additional `-log [PATH]` flag pointing to the directory you want the logs written to.
+
 See https://github.com/sillsdev/silnlp/issues/472 for original context.
 """
 
 import argparse
 import csv
+import logging
 import os
 import time
 
@@ -44,6 +47,9 @@ from dataclasses import dataclass
 from enum import Enum
 from pathlib import Path
 from typing import List, Optional
+
+
+logger = logging.getLogger(__name__)
 
 
 class Split(Enum):
@@ -67,9 +73,12 @@ class CliInput:
     target_iso: str
     dataset_descriptor: str
     output: Optional[str]
+    log: Optional[str]
 
 
 def load_sentence_pairs(input_file_path: str) -> List[SentencePair]:
+    logger.info("Loading sentence pairs")
+    logger.debug("Opening file")
     with open(input_file_path, "r", encoding="utf-8") as f:
         reader = csv.reader(f, delimiter="\t", quotechar='"')
         rows = list(reader)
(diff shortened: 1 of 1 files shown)


ddaspit (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/514)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @rminsil)

___
*[`silnlp/common/extract_xri.py` line 52 at r1](https://reviewable.io/reviews/sillsdev/silnlp/514#-O6c-Y0rDnV7vIpsFF4w:-O6c-Y0rDnV7vIpsFF4x:b-hdgdu1) ([raw file](https://github.com/sillsdev/silnlp/blob/7b9f514578460360ceabd3f9824f13d0780d6b00/silnlp/common/extract_xri.py#L52)):*
> ```Python
> 
> 
> logger = logging.getLogger(__name__)
> ```

You don't want to use `__name__` here, because it is set to `__main__`. You will want to use the following:

``` py
LOGGER = logging.getLogger(__package__ + ".extract_xri")
```
___
*[`silnlp/common/extract_xri.py` line 197 at r1](https://reviewable.io/reviews/sillsdev/silnlp/514#-O6c0FrMExsKaIQDlfBI:-O6c0FrMExsKaIQDlfBJ:bdmd8rx) ([raw file](https://github.com/sillsdev/silnlp/blob/7b9f514578460360ceabd3f9824f13d0780d6b00/silnlp/common/extract_xri.py#L197)):*
> ```Python
>         "-output", help="Optional path to the output directory where extract files are generated", type=str
>     )
>     parser.add_argument("-log", help="Optional path to enable logging. Logs are sent to the file passed", type=str)
> ```

In silnlp, we always have logging enabled and we output directly to stdout. I don't think there is a need for logging to a file.

rminsil:
> In silnlp, we always have logging enabled and we output directly to stdout. I don't think there is a need for logging to a file.

Each line of the file produces some logs and the files are around 4K entires, so you're typically getting about 8K-12K lines of logging, that's why I thought it better to have it off by default and send it to a file if it's enabled. Otherwise it's like a snowstorm in the terminal and it will usually overrun the terminal's buffer. We can pipe it to a file I guess but that's an extra step.

I was going to reserve stdout or stderr for the more meaningful output like statistics and summaries of data quality issues noticed.

Let me know what you think, I'm happy to do whatever.

rminsil:
@ddaspit I found the logging config in the parent __init.py__ and understand what you mean now.

I've removed the option to log to a file.

I've changed the logger name to be specific to the module. I'll make an analogous change to the repair logger used in #515.

I've added an optional flag to allow setting the debug level from the command line.

ddaspit (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/514#-:-O6hH2pw2fsVIHZTEnVK:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @rminsil)

PR #516, opened 2024-09-12
Check that necessary files and folders exist after arguments are parsed.

davidbaines:
Check that necessary files and folders exist after arguments are parsed. This ensures that necessary files exist before we create a job on clearml and gives the user more immediate feedback when there are typos in the command line.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/516)

### DIFF
--- a/silnlp/nmt/experiment.py
+++ b/silnlp/nmt/experiment.py
@@ -50,13 +50,8 @@ class SILExperiment:
             self.translate()
 
     def preprocess(self):
-        # Do some basic checks before starting the experiment
         exp_dir = Path(get_mt_exp_dir(self.name))
-        if not exp_dir.exists():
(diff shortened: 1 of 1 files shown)


isaac091:
For the translate script, it would probably be good to not check for the source project if `args.src_project` is `None`. When no value is passed for the source project, it defaults to the one in the config file, but the name of that project won't be retrieved until after the ClearML Task is created.

isaac091:
I'm also not sure that you would be able to check for an experiment or project's existence before attempting to copy it from the bucket, at least not when using the S3 bucket directly rather than via a local mounted drive.

davidbaines:
I think you're right Isaac. I'll have to rethink this.

Issue #329, opened 2024-02-22
When translating on a USFM file with a table spanning multiple verses, the output USFM drops some verse markers and table formatting markers 


mmartin9684-sil:
When the _translate_ script is run on a USFM file with a table that spans multiple verses, the output USFM file can omit mid-table verse markup as well as some of the table formatting markers.

As an example, the first attached file shows the USFM source for the NIV11R's version of NEH 7:7-39.  A single table spans verses 7-38, and a second table begins in verse 39.  The second attached file shows the translated USFM generated by the _translate_ script, in which the verse markers for verses 8-37 are missing ("\v 7" through "\v 38"); in addition, the table format markers "\tc1" and "\tcr2" are missing from the output.  The output USFM doesn't include verse markers until verse 39, in which the next table starts.
[NIV11R (NEH 7.7-39).txt](https://github.com/sillsdev/silnlp/files/14378131/NIV11R.NEH.7.7-39.txt)
[Mashi Draft (NEH 7.7-39).txt](https://github.com/sillsdev/silnlp/files/14378132/Mashi.Draft.NEH.7.7-39.txt)

ddaspit:
As a part of this issue, we should move to Machine for USFM parsing and generation. I believe that Machine supports table markers better.

mmartin9684-sil:
Would there be any other types of USFM markers where it would be helpful for us to double-check the way the markers are being handled in the drafts? Maybe we could do a little survey of what we're seeing in the drafts for markers where you suspect there could be problems.

ddaspit:
The least tested USFM markers would probably be [tables](https://ubsicap.github.io/usfm/tables/index.html), [milestones](https://ubsicap.github.io/usfm/milestones/index.html), [spacing and break markers](https://ubsicap.github.io/usfm/characters/index.html#spacing-and-breaks), and study Bible markers, such as [sidebars](https://ubsicap.github.io/usfm/notes_study/sidebars.html). Machine should have better support for most of these markers.

isaac091:
Using the machine.py parser, tables are handled much better, at least for these examples. See the attached file. All of the markers are preserved, though like in other situations where there are multiple markers in a verse, all of the pieces of the translated verse are inserted into the first text element which is directly after the verse marker. The "~~" bits are still missing, but I suspect that's related to the normalization done by the NLLB tokenizer, akin to the issues in #297.

[16NEH7.txt](https://github.com/user-attachments/files/17037483/16NEH7.txt)

PR #519, opened 2024-09-13
Use machine.py USFM parsing

### LINKED ISSUE #449: Update to use Machine.py for USFM generation
There are a number of issues with USFM generation in the translate.py script:
- #329
- #444 
- #379 
- #367 
- #330 
- #329 
- #158 

Most of these issues would probably be fixed if we scrapped the USFM parser that we currently use in the translate.py script and use the `UsfmTextUpdater` from Machine.py.

isaac091:
Overall, the new parsing is doing a much better job of preserving USFM structure and it's a lot easier to debug when there are differences.

I wasn't able to test the exact situations from each related issue, but all of the errors related to the initial parsing of files seem to be gone. I will follow up individually on some of the issues that are more unique, e.g. #329.

I also updated `paratext.py` and `check_books.py` to use `machine.py`'s functions whenever possible. Let me know if there are any other scripts that come to mind that could be cleaned up.

One additional change I made that's not directly related to the parser is that it no longer defaults to the target project in the config file if a target project is able to be used but one is not provided. I implemented it this way originally to match how the source project is treated, but ultimately I think it's just confusing to have the translation go into a different USFM structure when you don't specifically ask for it.

A couple of differences between the two parsers that I noticed, for future reference:
* Note that these observations are based on a small number of pairs that ultimately didn't have huge differences in structure
* Inserting translations into a different USFM structure: Overall, the new parser seems to be more consistent and does a better job of preserving the USFM structure of the target file. Both parsers only insert translated text when the source and target markers line up, but they handle mismatches differently.
  * When there is an inline element that exists only in the target structure, the old parser keeps it in the output, including the original text, but the new parser does not include the element in the output at all.
  * Conversely, for whole rows that don't match up, the new parser seems to preserve all of the markers of the target USFM and when there is no corresponding element in the source USFM, the original text from the target is also kept. In the same scenario, the old parser would keep the target markers but with no text or would not include the marker at all.
* The old parser was inconsistent about trying to translate empty verses vs leaving them alone. The new parser seems to always include the empty verses in the list of sentences to translate, so I changed the code to set them aside until translation is done.

Closes #449

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/519)


ddaspit (changes requested):
You should be able to remove all of the files in `silnlp/sfm`.

> Reviewed 7 of 7 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/519#-:-O6hKKRzDEcx1XmBPRdD:b-xlu47f)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @mshannon-sil)

ddaspit (approved):
:lgtm:

> Reviewed 7 of 7 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/519#-:-O6hND6A7HEh5cEYkiwT:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @mshannon-sil)

davidbaines:
On this branch, when trying to run extract_corpora I get this error:
```
F:\GitHub\silnlp>poetry run python -m silnlp.common.extract_corpora <project>
2024-09-16 15:48:27,672 - silnlp.common.environment - INFO - Using workspace: S:/ as per environment variable SIL_NLP_DATA_PATH.
2024-09-16 15:48:27,703 - silnlp.common.extract_corpora - INFO - Extracting <project>...
Traceback (most recent call last):
  File "C:\Python38\lib\runpy.py", line 194, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "C:\Python38\lib\runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "F:\GitHub\silnlp\silnlp\common\extract_corpora.py", line 105, in <module>
    main()
  File "F:\GitHub\silnlp\silnlp\common\extract_corpora.py", line 78, in main
    check_versification(project_dir)
  File "F:\GitHub\silnlp\silnlp\common\paratext.py", line 504, in check_versification
    project_versification: VersificationType = VersificationType(int(settings.versification))
TypeError: int() argument must be a string, a bytes-like object or a number, not 'Versification'
```
This error doesn't occur on the master branch.

ddaspit (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/519)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @mshannon-sil)

PR #580, opened 2024-10-31
Try to use a different iso code from the config when there is no test set

### LINKED ISSUE #578: Translating with no test set can result in drafts in the wrong language
With the new learning rate schedule and much faster training runs, it's possible to do onboarding with 2 training runs - a first run with typical train/val/test splits to assess the model quality, and a second run with only a train split to bump the model quality by using all available data for training (similar to Serval).  However, on this second model, if the experiment also uses the "--translate" option, the drafts can sometimes include content in the wrong target language.  For example, some parts of a Javanese draft were translated into English.

It appears that the translate step may rely on the test set to understand the target language, and can't specify it properly if there is no test set.

A workaround is to explicitly set the trg_iso in the translate_config.yml:
`translate: 
    books: MAT
    src_project: NIV11R
    checkpoint: last
    trg_iso: jav_Latn`

isaac091:
I opted for this approach over only throwing an error or a warning because in the case of translation, we're not really looking for language codes from the test set specifically, though I do think it still makes sense to look for a test code first because they will almost definitely match the languages we want to translate between.

Closes #578

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/580)

### DIFF
--- a/silnlp/nmt/translate.py
+++ b/silnlp/nmt/translate.py
@@ -84,6 +84,10 @@ class TranslationTask:
 
         if trg_iso is None:
             trg_iso = config.default_test_trg_iso
+            if trg_iso == "" and len(config.trg_isos) > 0:
+                trg_iso = next(iter(config.trg_isos))
+        if trg_iso == "":
+            LOGGER.warning("No language code was set for the target language")
 
         output_dir = config.exp_dir / "infer" / step_str / src_project
         if not config.model_dir.exists():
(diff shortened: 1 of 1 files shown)


ddaspit (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/580#-:-OAZSvhV3xbNC0KVms9i:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @isaac091)

PR #592, opened 2024-11-13
Clean up use of config.model

### LINKED ISSUE #590: Investigate repercussions of no longer having a default model name
Previously, configs always had a model name, as it defaulted to "SILTransformerBase". Now that OpenNMT is removed, if a model is not given, trying to access the model name gives a KeyError. There was a bug in `analyze_project_pairs` related to this that was already fixed.

I don't think there are any major issues this causes, but we may want to tweak how we look at model names.

isaac091:
The lack of a default value for config.model didn't end up affecting much, but I still altered it so I could clean up the code I previously changed.

I added a warning to experiment.py to help clarify errors coming from having a None model. I decided against a full-blown error because I think there are some situations where the name of the model technically isn't needed, i.e. when you already have a saved model + tokenizer. That being said, the average user will probably never have a trained model and an incomplete config file, so I can upgrade it to an error for simplicity if desired.

Closes #590

### DIFF
--- a/silnlp/nmt/analyze_project_pairs.py
+++ b/silnlp/nmt/analyze_project_pairs.py
@@ -106,12 +106,8 @@ def get_corpus_stats(config: Config, force_align: bool = False, deutero: bool = 
 
             src_script = predict_script_code("".join(corpus["source"][: min(len(corpus["source"]), 3000)]))
             trg_script = predict_script_code("".join(corpus["target"][: min(len(corpus["target"]), 3000)]))
-            try:
-                src_script_in_model = is_represented(src_script, config.model)
-                trg_script_in_model = is_represented(trg_script, config.model)
-            except:
-                src_script_in_model = None
-                trg_script_in_model = None
+            src_script_in_model = is_represented(src_script, config.model) if config.model is not None else None
+            trg_script_in_model = is_represented(trg_script, config.model) if config.model is not None else None
 
             stats_df.loc[project_pair, :] = [
                 pair_count,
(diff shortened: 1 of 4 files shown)


ddaspit (changes requested):
Would it be easier to just have a default value for model, such as NLLB 1.3B?

> Reviewed 4 of 4 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/592#-:-OBfL85a1pwZpZN1v6NZ:bw99vjp)* status: all files reviewed, 1 unresolved discussion (waiting on @isaac091)

___
*[`silnlp/nmt/config.py` line 432 at r1](https://reviewable.io/reviews/sillsdev/silnlp/592#-OBfL-5y3Ce8n4f7c5kp:-OBfL-5y3Ce8n4f7c5kq:b-9vgz8s) ([raw file](https://github.com/sillsdev/silnlp/blob/aa827e2758566c560e0806d6eec7d1193c0aee46/silnlp/nmt/config.py#L432)):*
> ```Python
> 
>     @property
>     def model(self) -> str:
> ```

The return type should be optional.

ddaspit (approved):
:lgtm:

> Reviewed 4 of 4 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/592#-:-OBftqGR8i2lrkWQNvAE:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @isaac091)

PR #600, opened 2024-11-27
Update argument handling to match other scripts.

davidbaines:
Update argument handling to match other scripts so that we don't specify --output-folder <experiment>.
Most other scripts take this directly as a required argument.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/600)


isaac091 (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/600)* status: all files reviewed, 2 unresolved discussions (waiting on @davidbaines)

___
*[`silnlp/common/collect_verse_counts.py` line 257 at r1](https://reviewable.io/reviews/sillsdev/silnlp/600#-ODgbfM26AKVDdfSlORZ:-ODgbfM26AKVDdfSlOR_:b-3s29lv) ([raw file](https://github.com/sillsdev/silnlp/blob/9f3bed9c4c47d1db2f6ec8f056287df09585fe9a/silnlp/common/collect_verse_counts.py#L257)):*
> ```Python
>         "--input-folder", default=SIL_NLP_ENV.mt_scripture_dir, help="Folder with corpus of Bible extracts"
>     )
>     parser.add_argument("folder", help="Folder containing config.yml file and where results will be saved.")
> ```

It might be good to put this as the first argument so that it's separated from the arguments that need to be explicitly named.

For the description, we should specify that the folder needs to be an experiment folder if the user is wanting to use a config file to define the input files, i.e. it won't work with any folder that has a config.yml file, it needs to be in the MT/experiments directory.
___
*[`silnlp/common/collect_verse_counts.py` line 287 at r1](https://reviewable.io/reviews/sillsdev/silnlp/600#-ODgcgHoAj5kfQJBMZjQ:-ODgcgHoAj5kfQJBMZjR:b-ohza0y) ([raw file](https://github.com/sillsdev/silnlp/blob/9f3bed9c4c47d1db2f6ec8f056287df09585fe9a/silnlp/common/collect_verse_counts.py#L287)):*
> ```Python
>         if not folder.exists():
>             LOGGER.error(
>                 f"Folder {folder} does not exist. Please provide a folder containing a config.yml file or a list of files with the --files argument."
> ```

Same thing here about needing an experiment folder (if --files is not being used)

isaac091 (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/600#-:-OLtZtlYDFmJ0-x8zThU:bnfp4nl)* status: all files reviewed, 2 unresolved discussions (waiting on @davidbaines)

Issue #285, opened 2024-01-03
Multi-GPU support for training and inference


mshannon-sil:
In order to use larger models such as the 3.3B NLLB model, we would need to use two 40GB A100 GPUs. And for those using GPUs that have less memory than the A100, they would need multiple of them to have enough memory to run the 1.3B NLLB model. This means we'll need to add support for multi-GPU training and inference. The number of GPUs to run experiments with should be configurable, defaulting to one GPU.

PR #612, opened 2024-12-19
#285 multigpu

mshannon-sil:
This PR adds support for splitting a model across two GPUs during training. There was already a num_devices argument in experiment.py that defaulted to 1, but it wasn't being used by any other part of the code, so I repurposed the argument. Now when the user specifies num_devices = 2 for an NLLB model, the model is loaded using the custom device map splitting it across two GPUs during the training phase.

I tested on 3 languages comparing 1 to 2 GPUs, and the BLEU scores were all within one point, with the 2 GPUs actually having a marginally higher BLEU score. These tests were conducted with the 40GB GPUs on the AQuA Server using the 1.3B NLLB model. I also started another set of 3 languages with the 2 GPU setup running on ORU, and although I confirmed that 2 24GB GPUs are now able to train the 1.3B NLLB when they couldn't before, the ORU server went down toward the end of training so I haven't been able to get BLEU scores yet.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/612)

### DIFF
--- a/silnlp/nmt/hugging_face_config.py
+++ b/silnlp/nmt/hugging_face_config.py
@@ -18,6 +18,7 @@ import pandas as pd
 import torch
 import transformers.utils.logging as transformers_logging
 import yaml
+from accelerate import infer_auto_device_map, init_empty_weights
 from accelerate.utils.memory import should_reduce_batch_size
 from datasets import Dataset
 from machine.scripture import ORIGINAL_VERSIFICATION, VerseRef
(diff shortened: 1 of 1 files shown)


ddaspit (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/612#-:-OFgz1dR-4rYLMLCCx2t:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @mshannon-sil)

PR #621, opened 2025-01-08
Add lexicon tools for use with XRI datasets

bhartmoore:
These two scripts run checks on XRI datasets. EITL team uses them regularly and I'd like to add them to silnlp.

- `count_words.py` prints total and unique word counts for the source and target datasets and gives stats on the diversity of target renderings for the most common source words. (May need a more precise name.)

- `compare_lex.py` compares lexicon data for two different source-target sets. It can be used to find what percent coverage each dataset has for the other, e.g. what percent of Bible words are represented in the training data. It defines "word" naively based on whitespace. Can be run on source side, target side, or both.

I've cleaned these up per a few helpful suggestions from Matthew: 

- lines < 120 chars
- uses SIL_NLP_ENV for paths to remain platform agnostic
- placed in silnlp/common

I expect there maybe concerns with style or functionality. In particular, `count_words.py` uses numpy, pandas, and collections.Counter, which may be cumbersome, and takes a long time to run. I'm glad to address all comments. Thanks.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/621)

### DIFF
--- a/silnlp/common/compare_lex.py
+++ b/silnlp/common/compare_lex.py
@@ -1,10 +1,11 @@
 import argparse
 import numpy
 import re
+from typing import List
 
 from ..common.environment import SIL_NLP_ENV
 
-def get_all_words(src_file: str) -> list:
+def get_all_words(src_file: str) -> List:
     words = []
     pattern = re.compile(r",(?=\S)")  # Look for commas with no following space
     with open(src_file, "r", encoding = "utf8") as src_data_file:     
@@ -18,7 +19,7 @@ def get_all_words(src_file: str) -> list:
                     words.append(word)  
     return words
 
-def find_unique(words1: list, words2: list) -> list:
+def find_unique(words1: List, words2: List) -> List:
     unique_words = []
     for word in words1:
         if word not in words2:
--- a/silnlp/common/compare_lex.py
+++ b/silnlp/common/compare_lex.py
@@ -1,0 +1,155 @@
+import argparse
+import numpy
+import re
+
+from ..common.environment import SIL_NLP_ENV
+
+def get_all_words(src_file: str) -> list:
+    words = []
+    pattern = re.compile(r",(?=\S)")  # Look for commas with no following space
+    with open(src_file, "r", encoding = "utf8") as src_data_file:     
+        for line in src_data_file:
+            for word in line.split(" "):
+                word = word.strip().strip("\'\"\\;,:.!?()-[]0123456789").lower()
+                finder = pattern.search(word)
+                if finder:             # Add space after commas as needed
+                    word = word[:finder.span()[1]]+" "+word[finder.span()[1]:]
+                if word != "":
+                    words.append(word)  
+    return words
+
+def find_unique(words1: list, words2: list) -> list:
+    unique_words = []
+    for word in words1:
+        if word not in words2:
+            unique_words.append(word)
+    return unique_words
+
+
+def main() -> None:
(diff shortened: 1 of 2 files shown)


mshannon-sil (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/621)* status: all files reviewed, 2 unresolved discussions (waiting on @bhartmoore and @ddaspit)

___
*[`silnlp/common/count_words.py` line 11 at r1](https://reviewable.io/reviews/sillsdev/silnlp/621#-OGAmQ2o3DdBlApYfUgE:-OGAmQ2o3DdBlApYfUgF:b-sdk2b0) ([raw file](https://github.com/sillsdev/silnlp/blob/9ff05e70db2ca524cf9c83824f8eb0906677860c/silnlp/common/count_words.py#L11)):*
> ```Python
> 
> def is_word(word: str) -> bool:
>     val = False
> ```

It looks this function defines a string as a word as long as it has a single alphabetic character. If that's the intended functionality, it can stop once any alphabetic character is found, to be more efficient, e.g.



_Code snippet:_
```Python
return any(char.isalpha() for char in word)
```
___
*[`silnlp/common/compare_lex.py` line 7 at r1](https://reviewable.io/reviews/sillsdev/silnlp/621#-OGAi---8YJHjnW2-0dM:-OGAi--02qm4Yod5c-No:bgjpnv9) ([raw file](https://github.com/sillsdev/silnlp/blob/9ff05e70db2ca524cf9c83824f8eb0906677860c/silnlp/common/compare_lex.py#L7)):*
> ```Python
> from ..common.environment import SIL_NLP_ENV
> 
> def get_all_words(src_file: str) -> list:
> ```

The `List` type from `from typing import List` should be used for consistency with the rest of the repo

bhartmoore:
@mshannon-sil Thank you for the review. I've made changes to address your comments. 
Should `is_word()` still be a function now that it's a single line? It's only called once. 
You're right about the functionality; it's meant to get rid of numbers and stray punctuation that were showing up in the lexicon.

mshannon-sil:
I missed that it was only called once. In that case, I wouldn't keep it as a function.

bhartmoore:
> I missed that it was only called once. In that case, I wouldn't keep it as a function.

Looking back, I realized that it was because we may want to update the definition of a word over time (e.g. in the newest set, there are alpha words in the source that are translated as numerals in the target, so I probably need to allow for those), and we may want to make it more sophisticated than just looking at whitespace. So if you're ok with it, I may keep it as a function even though it's reduced to one (very elegant!) line.

benjaminking:
I think it could be worthwhile to keep the single-line `is_word` function here, because the name of the function can act as a bit of documentation, especially since the purpose of that line might not be immediately obvious otherwise.

mshannon-sil (commented):
Makes sense, it's fine to keep it as is then.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/621#-:-OGFscLl7bMXFYo6r3O3:b-v4j5nb)* status: 0 of 2 files reviewed, 2 unresolved discussions (waiting on @bhartmoore and @ddaspit)

mshannon-sil (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/621#-:-OGFsr4yEt2a_4Z7nmHY:bnfp4nl)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @bhartmoore and @ddaspit)

PR #624, opened 2025-01-09
S3 bucket script

benjaminking:
This PR modifies the previous S3 bucket script to support automatic mounting of the S3 bucket in a dev container for both WSL2-based Ubuntu systems and standalone Ubuntu systems.  It splits the logic of the previous script into two portions:

* the part that does not depend on any SILNLP repo files (this logic has been moved into the Dockerfile) and
* the part that uses files in the SILNLP repo (this remains in the `s3_bucket_setup.sh` script, which is run as a post-install command)

The contents of the S3 bucket should now be accessible at `/silnlp` upon creation of the dev container.  The major change required to allow this to run on standalone Ubuntu systems is to run the Docker container with a different AppArmor profile.  For simplicity, I configured the container to run in "unconfined" mode, but if more fine-grained security is desired (or if ClearML requires it), we should be able to create a custom AppArmor profile that we distribute in the SILNLP repo.

This has been tested on both WSL2-based Ubuntu (in Windows 11) and in standalone Linux Mint, a Ubuntu-based distro.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/624)


mshannon-sil (changes requested):
I tested on my Ubuntu machine, and it's also working there. I'll check with Bryan from TechOps to see if he has any security concerns with running the container like this.

> Reviewed 1 of 3 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/624#-:-OGHBhDqE__oTVcWc3JF:burpygx)* status: 1 of 3 files reviewed, 3 unresolved discussions (waiting on @benjaminking)

___
*[`.devcontainer/Dockerfile` line 26 at r1](https://reviewable.io/reviews/sillsdev/silnlp/624#-OGH9sSg2hT5CDwrg6nw:-OGH9sSg2hT5CDwrg6nx:bwszrqs) ([raw file](https://github.com/sillsdev/silnlp/blob/6535c743a88f8832c3ff98f0ed5eeee68f8035ee/.devcontainer/Dockerfile#L26)):*
> ```Dockerfile
>     curl \
>     unzip
> RUN apt-get update
> ```

I'm not sure it's necessary to call apt-get update again, since it's already called right before this, unless apt-get needs to be updated only after certain packages have been installed.
___
*[`.devcontainer/Dockerfile` line 28 at r1](https://reviewable.io/reviews/sillsdev/silnlp/624#-OGHAEZRDW4LGt6nmjKJ:-OGHAEZRDW4LGt6nmjKK:b-bsxyw9) ([raw file](https://github.com/sillsdev/silnlp/blob/6535c743a88f8832c3ff98f0ed5eeee68f8035ee/.devcontainer/Dockerfile#L28)):*
> ```Dockerfile
> RUN apt-get update
> RUN apt-get install --no-install-recommends -y \
>     less \
> ```

I don't see these two packages used anywhere. Are they just being added as helpful developer tools? I think they'd be useful, especially since I use nano quite a bit. I think it'd be cleaner to move this to the larger apt-get install section directly above it
___
*[`.devcontainer/Dockerfile` line 54 at r1](https://reviewable.io/reviews/sillsdev/silnlp/624#-OGHAfpp84IjLvRS6o1x:-OGHAfpp84IjLvRS6o1y:b-yciq1p) ([raw file](https://github.com/sillsdev/silnlp/blob/6535c743a88f8832c3ff98f0ed5eeee68f8035ee/.devcontainer/Dockerfile#L54)):*
> ```Dockerfile
> CMD ["bash"]
> # Set up the S3 bucket
> RUN apt update
> ```

Is this separated out because it needs to use apt instead of apt-get, or could it be added to the longer apt-get install earlier in the dockerfile?

benjaminking:
I updated the Dockerfile based on your comments.  Firstly, `less` and `nano` were included by mistake -- I had those in my Dockerfile for convenience.  (Strangely, to get `apt-get` to recognize them, you have to call `apt-get update` a second time)  But I've removed those for now.

Also, for reasons I don't understand, if you want to install `fuse3` and `rclone` on WSL2, they have to be in a separate `apt-get` command from the other installs.  I've moved them up to the same section as the other `apt-get` installs.  I also switched to using the bundled version of `rclone` instead of downloading it from its website, and that doesn't seem to cause any problems.

mshannon-sil (changes requested):
> Reviewed 1 of 3 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/624)* status: all files reviewed, 1 unresolved discussion (waiting on @benjaminking)

___
*[`.devcontainer/Dockerfile` line 48 at r2](https://reviewable.io/reviews/sillsdev/silnlp/624#-OGjxVzbEs36x_9oWhH6:-OGjxVzbEs36x_9oWhH7:b-2axfae) ([raw file](https://github.com/sillsdev/silnlp/blob/c8cc21cd61a0e33c58dcfb04740022e693ba9d8b/.devcontainer/Dockerfile#L48)):*
> ```Dockerfile
> ENV SIL_NLP_CACHE_PROJECT_DIR=/root/.cache/silnlp/projects
> # Set up the S3 bucket
> RUN mkdir -p /silnlp
> ```

To be consistent with the readme instructions, which mount at ~/S, it's best to mount it at /root/S. I double checked in my docker container, and the ~ is equivalent to the /root folder.

benjaminking:
I've just pushed a commit that changes the mount point to `~/S`, adds an AppArmor profile, and updates `devcontainer.json` to use the new AppArmor profile.

To register the new AppArmor profile, you need to run `sudo apparmor_parser -r -W docker-apparmor`.  This should only need to be done once for each machine.

I believe that Docker will ignore the AppArmor profile on systems that don't support AppArmor (or don't have it installed), so there shouldn't be any impact to non-Linux users.  The AppArmor profile is also the same as the default Docker profile, except that it allows the container to perform mounts.

ddaspit (commented):
> Reviewed 4 of 4 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/624)* status: all files reviewed, 1 unresolved discussion (waiting on @benjaminking)

ddaspit (commented):
LGTM, but I will let someone who has a better understanding of these things approve it.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/624#-:-OIWDDdRBGSzFQR8GG8h:b-wz7smk)* status: all files reviewed, 1 unresolved discussion (waiting on @benjaminking)

mshannon-sil (commented):
@benjaminking Sorry this was in my backlog to review for so long. I've tested it and it's working for me. Now that we have an app armor profile, would we like to go ahead and extend this not just for dev containers but for the AQuA/Cheetah servers as well? This might help with avoiding the failures for uploading checkpoints on those servers.

> Reviewed 3 of 4 files at r3, 1 of 1 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/624#-:-OL0bz6eEfvsqndT0rKU:b-7l84n1)* status: all files reviewed, 1 unresolved discussion (waiting on @benjaminking)

benjaminking:
That's ok.  It's been on my backlog as well.  I think the next step for us is to install the App Armor profile on one of our servers, modify the Docker setup in `clearml_connection.py` to use the new profile, and test that the container is still created normally.  If that works, we should be able to test connecting to the S3 bucket automatically.

mshannon-sil:
I actually already installed the app armor profile on the AQuA and Cheetah servers about an hour ago to prepare for testing on them, using the command you listed. So it should be all set for any changes you'd like to make to clearml_connection.py.

benjaminking:
Excellent.  I will try to test out creating a Docker container using the new profile.

benjaminking:
I was able to successfully create a Docker container with the new AppArmor profile and run a task with it.  My next step is to try to automatically connect to the S3 bucket from inside the container.

benjaminking:
It looks like my attempts at connecting to the S3 bucket from ClearML are failing because the `rclone.conf` file on the Docker image no longer has a `[s3silnlp]` section.  Have we removed support for S3 on ClearML already?

(There are workarounds I can probably do, but I wasn't sure what the status was regarding S3 and ClearML)

mshannon-sil:
@TaperChipmunk32 I think we were still supporting aws for a little while longer, so it might be good to add the section back to rclone.conf? Also I thought that IDX might be using their own aws S3 bucket with SILNLP, and if that's the case it would be good to leave the support in.

@benjaminking We are switching over to MinIO next week, and support for it was recently added to master, so it also might make sense to try adapting your changes to MinIO.

TaperChipmunk32:
> @TaperChipmunk32 I think we were still supporting aws for a little while longer, so it might be good to add the section back to rclone.conf? Also I thought that IDX might be using their own aws S3 bucket with SILNLP, and if that's the case it would be good to leave the support in.
> 
> @benjaminking We are switching over to MinIO next week, and support for it was recently added to master, so it also might make sense to try adapting your changes to MinIO.

Good point about IDX, we can leave AWS support in, I'll just need to update the warning message. I can also add back AWS to the rclone.conf file. I'll create a PR for both changes.

ddaspit (commented):
> Reviewed 1 of 1 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/624)* status: all files reviewed, 1 unresolved discussion (waiting on @benjaminking)

TaperChipmunk32:
@ddaspit I just realized something from our discussions on the `translate.py` PR. Once this is merged in, we could potentially require the bucket to be mounted. This would cause us to no longer need most of the S3 specific functions and libraries (`boto3`, `S3Path`, `copy_experiment_to_bucket`, etc.) and the cache system in silnlp, since rclone would handle all of the uploads/downloads. The exceptions would be `clean_s3.py` and `sync_minio_b2.py`

TaperChipmunk32:
@benjaminking I can add a commit to switch this PR over to using MinIO and B2 instead of AWS, if you would like me to.

benjaminking:
Yes, please feel free to change it over to using MinIO and Backblaze.

TaperChipmunk32:
> Yes, please feel free to change it over to using MinIO and Backblaze.

Support should be in for MinIO now.

benjaminking:
> That is a good point. I would love to be able to strip out all of the S3 code. We should discuss that option further.

I am looking again into alternative ways to automatically mount the bucket on the AQuA servers, which is the last major hurdle before we can merge this in.

benjaminking:
I was able to get the AQuA machines to automatically connect to the MinIO bucket on ClearML jobs.  We should be able to look into merging this soon, if everyone is comfortable with the way I implemented this.

mshannon-sil (approved):
:lgtm:

> Reviewed 2 of 2 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/624#-:-OMwvacp76ZJR-9_phuK:bnfp4nl)* status: all files reviewed, 2 unresolved discussions (waiting on @benjaminking)

Issue #623, opened 2025-01-08
Use a different experiment folder's checkpoints as the base model when running an silnlp experiment on clearml


laura-burdick-sil:
Enable multi-stage fine-tuning of NLLB, by setting the "model" parameter in the config file to a previously trained model, rather than the base huggingface model.

Damien (slack): "Our old OpenNMT implementation supported parent models and multi-stage finetuning. I guess we never implemented it for Huggingface. You can even see that there is a `has_parent` property on the `Config` class that isn't used for anything after we removed OpenNMT support. You would configure it by setting the `data/parent` setting to the experiment name. Huggingface transformers definitely supports fine tuning from a local checkpoint, so we should be able to implement it."

PR #628, opened 2025-01-15
Update for Issue #623 

AmeWenJ:
Update the `hugging_face_config.py`:

- Load the model from the parent experiment if a parent is specified in the config. 
- The tokenizer will be the same as the parent experiment unless it is specified in the config.
- An optional parameter is added in `SilSeq2SeqTrainer` to store the model prefix (when the model is from a parent experiment, the model name will be a path, so we can't get directly a prefix).

This modification has been tested on:

- [x] Training without parent or LoRA.
- [x] Training without parent but with LoRA.
- [x] Training with parent and LoRA.
- [x] Training with parent but without LoRA.

The remaining potential issues:

- Model name becomes a path when the parent is specified, so it is not safe to directly use it anymore (such as in methods that need to get a prefix from the model name). That is why I edited the trainer class. I'm not sure if there are other methods that will be influenced, but I went over the related methods and think it's all good now.

- If the parent model prefix is different from the model specified in the config, it will throw an exception. Based on my intuition, since we are supposed to use the parent model, the model specified in the config should be ignored? It feels weird that it is specified, so I see it as a way to double check something like "This is the model I want! It should be an NLLB instead of T5". 

Please let me know if there is any further update or modification needed.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/628)

### DIFF
--- a/silnlp/nmt/experiment.py
+++ b/silnlp/nmt/experiment.py
@@ -58,13 +58,16 @@ class SILExperiment:
         if not config_file.exists():
             raise RuntimeError(f"ERROR: Config file does not exist in experiment folder {exp_dir}.")
         SIL_NLP_ENV.copy_experiment_from_bucket(self.name)
+        if self.config.has_parent:
+            SIL_NLP_ENV.copy_experiment_from_bucket(self.config.data["parent"])
         self.config.preprocess(self.make_stats, self.force_align)
         SIL_NLP_ENV.copy_experiment_to_bucket(self.name, overwrite=self.force_align)
 
     def train(self):
         os.system("nvidia-smi")
         SIL_NLP_ENV.copy_experiment_from_bucket(self.name)
-
+        if self.config.has_parent:
+            SIL_NLP_ENV.copy_experiment_from_bucket(self.config.data["parent"])
         model = self.config.create_model(self.mixed_precision, self.num_devices)
         model.save_effective_config(self.config.exp_dir / f"effective-config-{self.rev_hash}.yml")
         SIL_NLP_ENV.copy_experiment_to_bucket(self.name, patterns=f"effective-config-{self.rev_hash}.yml")
--- a/silnlp/nmt/hugging_face_config.py
+++ b/silnlp/nmt/hugging_face_config.py
@@ -194,14 +194,9 @@ def load_parent_exp(parent_exp: str) -> Tuple[str, str]:
     if not parent_model_dir.is_dir():
         LOGGER.error("The parent experiment does not contain a checkpoint.")
         raise ValueError(f"Invalid path {parent_model_dir}")
-
-    trainer_state_path = parent_model_dir / "trainer_state.json"
-    with open(trainer_state_path, "r") as file:
-        trainer_state = json.load(file)
-    parent_model_path = trainer_state.get("best_model_checkpoint", None)
-    parent_model_name = os.path.basename(parent_model_path)
-    parent_model = parent_model_dir / parent_model_name
-
+    parent_model = get_last_checkpoint(parent_model_dir)
+    if has_best_checkpoint(parent_model_dir):
+        parent_model = get_best_checkpoint(parent_model_dir)
     parent_model_files = os.listdir(parent_model)
     valid_model_files = [f for f in parent_model_files if f.endswith(".safetensors")]
     if not valid_model_files:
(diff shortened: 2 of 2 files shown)


isaac091 (commented):
My current understanding of how the tokenizer is chosen is as follows (assuming a new experiment):
1. If the config file of the experiment says to update the tokenizer, initialize a new tokenizer to be updated.
2. Else if there is a parent model, use the tokenizer from the parent.
3. Otherwise initialize a new tokenizer

Is that correct? I don't immediately have an opinion about this, but I'm curious if there was a discussion around when to use the parent tokenizer vs a new one?

> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/628#-:-OGfwbtCBAn8U99ot9Tv:b-4t9sb3)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @ddaspit)

AmeWenJ:
> My current understanding of how the tokenizer is chosen is as follows (assuming a new experiment):
> 
> 1. If the config file of the experiment says to update the tokenizer, initialize a new tokenizer to be updated.
> 2. Else if there is a parent model, use the tokenizer from the parent.
> 3. Otherwise initialize a new tokenizer
> 
> Is that correct? I don't immediately have an opinion about this, but I'm curious if there was a discussion around when to use the parent tokenizer vs a new one?
> 
> > Reviewed 1 of 1 files at r1, all commit messages.
> > _[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/628#-:-OGfwbtCBAn8U99ot9Tv:b-4t9sb3)_ status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @ddaspit)


Thank you for bringing it up! And yes, your understanding is correct. Since this will be used for Laura's project, I'm not sure if she will prefer to specify a tokenizer. That's why I keep the option of specifying a tokenizer in the config file. However, in my testing, I didn't specify it, so all tokenizers were from the parent directory. 

I'll edit my summary a bit. Thank you!

laura-burdick-sil:
Hmm, that's a good question. I think that most of the time it makes sense to use the tokenizer from the parent model (since the parent model has already been trained with that tokenizer). Maybe it's worth keeping the option to specify another tokenizer, if that's the desired behavior, though?

AmeWenJ:
> Hmm, that's a good question. I think that most of the time it makes sense to use the tokenizer from the parent model (since the parent model has already been trained with that tokenizer). Maybe it's worth keeping the option to specify another tokenizer, if that's the desired behavior, though?

Hi @laura-burdick-sil, I "designed" it in this way because I was not 100% sure about your project, so I prioritize config in case you may need to use a specific tokenizer or different source of languages. However, I personally would prefer to prioritize parent tokenizer because as you said, the model was trained with it. If you think it is better to just use the parent tokenizer all the time, I can change it. It will just be one line of code so don't worry!

ddaspit (approved):
I think we can move forward with what you currently have. I think it is logical to allow you to specify a new tokenizer to override the parent tokenizer. We can always change it later, if we find that it doesn't work for us. This is excellent work.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/628#-:-OGl2cc5AyZkIBW-vduK:b-l9c78r)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @AmeWenJ)

AmeWenJ:
Thank you for all your reviews! I just made another tiny change in the checkpoint loading part. Now it prioritizes loading the last checkpoint instead of the best one. I've also added @laura-burdick-sil as a reviewer. Once she believes this update can support the project going on, it should be good to merge!

ddaspit (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/628)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @laura-burdick-sil)

AmeWenJ:
The previous version downloads the whole parent folder during the config initialization, which is slow and consuming local storage. To improve the efficiency, I made the following changes:

- Method `get_parent_model_prefix` is a part of the previous `load_parent_exp`. It now serves for the purpose of downloading the `config.yml` file from the parent experiment folder and return the parent model prefix.
- Method `get_parent_model_name` is the other part of the previous `load_parent_exp`. It now serves for the purpose of downloading the `run/trainer_state.json` file from the parent experiment folder and return the parent model path.
- Method `get_parent_last_checkpoint` is a helper method for `get_parent_model_name`, because the previous `get_last_checkpoint` can only be used when model files actually exist.
- Download the remaining files from the parent experiment at the beginning of preprocessing and training. 

The potential risk is that `get_parent_last_checkpoint` assumes there is a model checkpoint at the `max_steps`, which is usually the case and I think it should be like that - After all, it is supposed to be the last step of training, which is the last version of the model.


Based on the feedback, this should be the last problem, so if everything goes well, it should finally be good to merge...!

laura-burdick-sil (approved):
Great work, Wenfan! Thank you for making the changes so the model isn't downloaded locally each time.

ddaspit (commented):
> Reviewed 2 of 2 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/628)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @AmeWenJ)

PR #637, opened 2025-01-25
Fix tokenizer bug when resumes from a parent experiment

AmeWenJ:
I forgot to `get_mt_exp_dir()` before sending the parent path to the tokenizer :(

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/637)


laura-burdick-sil (approved):
Thanks for fixing this!

PR #651, opened 2025-02-11
Multi Target Support

TaperChipmunk32:
-Added support for Multi-Target training and testing
-Made many_to_many mapping the default for corpus pairs to prevent crashes

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/651)

### DIFF
--- a/silnlp/nmt/config.py
+++ b/silnlp/nmt/config.py
@@ -232,7 +232,7 @@ def parse_corpus_pairs(corpus_pairs: List[dict]) -> List[CorpusPair]:
         is_lexical_data: bool = pair["lexical"]
 
         if "mapping" not in pair:
-            pair["mapping"] = DataFileMapping.ONE_TO_ONE.name.lower()
+            pair["mapping"] = DataFileMapping.MANY_TO_MANY.name.lower()
         mapping = DataFileMapping[pair["mapping"].upper()]
         if not is_scripture and mapping != DataFileMapping.ONE_TO_ONE:
             raise RuntimeError("Basic corpus pairs only support one-to-one mapping.")
--- a/silnlp/nmt/config.py
+++ b/silnlp/nmt/config.py
@@ -320,6 +320,7 @@ class NMTModel(ABC):
         self,
         input_paths: List[Path],
         translation_paths: List[Path],
+        src_trg_isos: List[Tuple[str, str]],
         produce_multiple_translations: bool = False,
         vref_paths: Optional[List[Path]] = None,
         ckpt: Union[CheckpointType, str, int] = CheckpointType.LAST,
--- a/silnlp/nmt/config.py
+++ b/silnlp/nmt/config.py
@@ -232,7 +232,7 @@ def parse_corpus_pairs(corpus_pairs: List[dict]) -> List[CorpusPair]:
         is_lexical_data: bool = pair["lexical"]
 
         if "mapping" not in pair:
-            pair["mapping"] = DataFileMapping.MANY_TO_MANY.name.lower()
+            pair["mapping"] = DataFileMapping.ONE_TO_ONE.name.lower()
         mapping = DataFileMapping[pair["mapping"].upper()]
         if not is_scripture and mapping != DataFileMapping.ONE_TO_ONE:
             raise RuntimeError("Basic corpus pairs only support one-to-one mapping.")
--- a/silnlp/nmt/hugging_face_config.py
+++ b/silnlp/nmt/hugging_face_config.py
@@ -1088,25 +1088,28 @@ class HuggingFaceNMTModel(NMTModel):
         self,
         input_paths: List[Path],
         translation_paths: List[Path],
+        src_trg_isos: List[Tuple[str, str]],
         produce_multiple_translations: bool = False,
         vref_paths: Optional[List[Path]] = None,
         ckpt: Union[CheckpointType, str, int] = CheckpointType.LAST,
     ) -> None:
+        lang_codes: Dict[str, str] = self._config.data["lang_codes"]
         tokenizer = self._config.get_tokenizer()
         model = self._create_inference_model(ckpt, tokenizer)
-        pipeline = PretokenizedTranslationPipeline(
-            model=model,
-            tokenizer=tokenizer,
-            src_lang=self._config.test_src_lang,
-            tgt_lang=self._config.test_trg_lang,
-            device=0,
-        )
-        pipeline.model = torch.compile(pipeline.model)
-        for input_path, translation_path, vref_path in zip(
+        for input_path, translation_path, src_trg_iso, vref_path in zip(
             input_paths,
             translation_paths,
+            src_trg_isos,
             cast(Iterable[Optional[Path]], repeat(None) if vref_paths is None else vref_paths),
         ):
+            pipeline = PretokenizedTranslationPipeline(
+                model=model,
+                tokenizer=tokenizer,
+                src_lang=lang_codes.get(src_trg_iso[0]),
+                tgt_lang=lang_codes.get(src_trg_iso[1]),
+                device=0,
+            )
+            pipeline.model = torch.compile(pipeline.model)
             length = count_lines(input_path)
             with ExitStack() as stack:
                 src_file = stack.enter_context(input_path.open("r", encoding="utf-8-sig"))
--- a/silnlp/nmt/hugging_face_config.py
+++ b/silnlp/nmt/hugging_face_config.py
@@ -1096,6 +1096,8 @@ class HuggingFaceNMTModel(NMTModel):
         lang_codes: Dict[str, str] = self._config.data["lang_codes"]
         tokenizer = self._config.get_tokenizer()
         model = self._create_inference_model(ckpt, tokenizer)
+        model.to(0)
+        model = torch.compile(model)
         for input_path, translation_path, src_trg_iso, vref_path in zip(
             input_paths,
             translation_paths,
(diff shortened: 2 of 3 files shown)


ddaspit (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/651)* status: all files reviewed, 3 unresolved discussions (waiting on @TaperChipmunk32)

___
*[a discussion](https://reviewable.io/reviews/sillsdev/silnlp/651#-OIpu_PD6rAW-IOb9K3s:-OIpu_PD6rAW-IOb9K3t:bb1geuc) (no related file):*
I wanted to determine if using multiple sources is broken or not. It looks like it is working correctly. Although we set `src_lang` on the pipeline, it isn't actually used. This is because the data is already tokenized and contains the source language code. You can see this in `PretokenizedTranslationPipeline`. The `tgt_lang` is used, since the `forced_bos_token_id` property needs to be set, so the decoder can properly force the first token to the target language code.
___
*[`silnlp/nmt/config.py` line 235 at r1](https://reviewable.io/reviews/sillsdev/silnlp/651#-OIpqMY08B0hVWTMSMfN:-OIpqMY08B0hVWTMSMfO:bjrb5o4) ([raw file](https://github.com/sillsdev/silnlp/blob/499f1a7f3b4df1306768fdbf20258bcdfb4a9676/silnlp/nmt/config.py#L235)):*
> ```Python
> 
>         if "mapping" not in pair:
>             pair["mapping"] = DataFileMapping.MANY_TO_MANY.name.lower()
> ```

Changing the defaults might break existing config files. Are we sure this is safe?
___
*[`silnlp/nmt/hugging_face_config.py` line 1112 at r1](https://reviewable.io/reviews/sillsdev/silnlp/651#-OIpqeOf4zk4fkvCCGwW:-OIpqeOg2sB-PKxDblN7:b9xgm5f) ([raw file](https://github.com/sillsdev/silnlp/blob/499f1a7f3b4df1306768fdbf20258bcdfb4a9676/silnlp/nmt/hugging_face_config.py#L1112)):*
> ```Python
>                 device=0,
>             )
>             pipeline.model = torch.compile(pipeline.model)
> ```

I feel like we shouldn't be calling `torch.compile` on the same model multiple times. I'm not sure if it will recompile every time or not. Could we try calling this once? You will probably need to move the model to the GPU first before calling `torch. Compile`, i.e.

```py
model.to(0)
```

TaperChipmunk32 (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/651)* status: all files reviewed, 3 unresolved discussions (waiting on @ddaspit)

___
*[a discussion](https://reviewable.io/reviews/sillsdev/silnlp/651#-OIpu_PD6rAW-IOb9K3s:-OIqK40R5fsImBXJKeLF:bkervld) (no related file):*

I wanted to determine if using multiple sources is broken or not. It looks like it is working correctly. Although we set `src_lang` on the pipeline, it isn't actually used. This is because the data is already tokenized and contains the source language code. You can see this in `PretokenizedTranslationPipeline`. The `tgt_lang` is used, since the `forced_bos_token_id` property needs to be set, so the decoder can properly force the first token to the target language code.
</blockquote></details>

Thanks for looking into that. This is my understanding of it as well now.
___
*[`silnlp/nmt/config.py` line 235 at r1](https://reviewable.io/reviews/sillsdev/silnlp/651#-OIpqMY08B0hVWTMSMfN:-OIqGJTv4_kRIof2ZG5z:b-bri0a9) ([raw file](https://github.com/sillsdev/silnlp/blob/499f1a7f3b4df1306768fdbf20258bcdfb4a9676/silnlp/nmt/config.py#L235)):*

Changing the defaults might break existing config files. Are we sure this is safe?
</blockquote></details>

It should not break any existing config files. In fact, `one_to_one` and `many_to_many` mappings could likely be removed. The `one_to_one` case is handled by `many_to_many`. In the future, we could remove `mapping` as a config option and replace it with `mixed_src` as a boolean value. The only time `many_to_many` versus `one_to_one` mapping is mentioned is in this section and in `get_data_file_pairs` from line 302. I believe both of these instances could be removed with some minor refactoring.



_Code snippet:_
```Python
def get_data_file_pairs(corpus_pair: CorpusPair) -> Iterable[Tuple[DataFile, DataFile]]:
    if corpus_pair.mapping == DataFileMapping.ONE_TO_ONE:
        for file_pair in zip(corpus_pair.src_files, corpus_pair.trg_files):
            yield file_pair
    else:
        for src_file in corpus_pair.src_files:
            for trg_file in corpus_pair.trg_files:
                yield (src_file, trg_file)
```
___
*[`silnlp/nmt/hugging_face_config.py` line 1112 at r1](https://reviewable.io/reviews/sillsdev/silnlp/651#-OIpqeOf4zk4fkvCCGwW:-OIqKP7f9K9ihjUmhKGe:b-96c3e0) ([raw file](https://github.com/sillsdev/silnlp/blob/499f1a7f3b4df1306768fdbf20258bcdfb4a9676/silnlp/nmt/hugging_face_config.py#L1112)):*

I feel like we shouldn't be calling `torch.compile` on the same model multiple times. I'm not sure if it will recompile every time or not. Could we try calling this once? You will probably need to move the model to the GPU first before calling `torch. Compile`, i.e.

```py
model.to(0)
```
</blockquote></details>

I will be looking into this.

TaperChipmunk32 (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/651)* status: 2 of 3 files reviewed, 3 unresolved discussions (waiting on @ddaspit)

___
*[`silnlp/nmt/hugging_face_config.py` line 1112 at r1](https://reviewable.io/reviews/sillsdev/silnlp/651#-OIpqeOf4zk4fkvCCGwW:-OIqs-8F2reUF-CoI7M1:b-tqms00) ([raw file](https://github.com/sillsdev/silnlp/blob/499f1a7f3b4df1306768fdbf20258bcdfb4a9676/silnlp/nmt/hugging_face_config.py#L1112)):*

I will be looking into this.
</blockquote></details>

With calling `torch.compile` once and passing the compiled model into `pipeline` an error message is thrown. However, it does not crash and testing continues as expected and produces the same results.



_Code snippet:_
```Python
[ERROR|base.py:1149] 2025-02-11 21:15:19,616 >> The model 'OptimizedModule' is not supported for . Supported models are ['BartForConditionalGeneration', 'BigBirdPegasusForConditionalGeneration', 'BlenderbotForConditionalGeneration', 'BlenderbotSmallForConditionalGeneration', 'EncoderDecoderModel', 'FSMTForConditionalGeneration', 'GPTSanJapaneseForConditionalGeneration', 'LEDForConditionalGeneration', 'LongT5ForConditionalGeneration', 'M2M100ForConditionalGeneration', 'MarianMTModel', 'MBartForConditionalGeneration', 'MT5ForConditionalGeneration', 'MvpForConditionalGeneration', 'NllbMoeForConditionalGeneration', 'PegasusForConditionalGeneration', 'PegasusXForConditionalGeneration', 'PLBartForConditionalGeneration', 'ProphetNetForConditionalGeneration', 'Qwen2AudioForConditionalGeneration', 'SeamlessM4TForTextToText', 'SeamlessM4Tv2ForTextToText', 'SwitchTransformersForConditionalGeneration', 'T5ForConditionalGeneration', 'UMT5ForConditionalGeneration', 'XLMProphetNetForConditionalGeneration'].
```

ddaspit (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/651)* status: all files reviewed, 1 unresolved discussion (waiting on @TaperChipmunk32)

___
*[`silnlp/nmt/config.py` line 235 at r1](https://reviewable.io/reviews/sillsdev/silnlp/651#-OIpqMY08B0hVWTMSMfN:-OIqvhxQ6nqRA-oxgvMs:ba8tf00) ([raw file](https://github.com/sillsdev/silnlp/blob/499f1a7f3b4df1306768fdbf20258bcdfb4a9676/silnlp/nmt/config.py#L235)):*

It should not break any existing config files. In fact, `one_to_one` and `many_to_many` mappings could likely be removed. The `one_to_one` case is handled by `many_to_many`. In the future, we could remove `mapping` as a config option and replace it with `mixed_src` as a boolean value. The only time `many_to_many` versus `one_to_one` mapping is mentioned is in this section and in `get_data_file_pairs` from line 302. I believe both of these instances could be removed with some minor refactoring.
</blockquote></details>

There is a difference between `one_to_one` and `many_to_many` if you specify multiple sources/targets. `one_to_one` maps the first source to the first target, the second source to the second target, and so on. `many_to_many` maps every source to every target. So, I would change the default back to `one_to_one`, since it could change the behavior for existing configs.

TaperChipmunk32 (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/651)* status: 2 of 3 files reviewed, 1 unresolved discussion (waiting on @ddaspit)

___
*[`silnlp/nmt/config.py` line 235 at r1](https://reviewable.io/reviews/sillsdev/silnlp/651#-OIpqMY08B0hVWTMSMfN:-OIqxj1E6598yFvTTPS4:b6u9lvd) ([raw file](https://github.com/sillsdev/silnlp/blob/499f1a7f3b4df1306768fdbf20258bcdfb4a9676/silnlp/nmt/config.py#L235)):*

There is a difference between `one_to_one` and `many_to_many` if you specify multiple sources/targets. `one_to_one` maps the first source to the first target, the second source to the second target, and so on. `many_to_many` maps every source to every target. So, I would change the default back to `one_to_one`, since it could change the behavior for existing configs.
</blockquote></details>

Oh, I see, I was misinterpreting `one_to_one` to mean only one source and one target. Thanks for the clarification.

ddaspit (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/651)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @TaperChipmunk32)

PR #678, opened 2025-03-14
Fixed an issue with interactive sessions

TaperChipmunk32:
There was an issue found with interactive sessions that caused them to never try to connect to an S3 bucket since `SIL_NLP_DATA_PATH=/silnlp` is included in the clearml.conf files and `BUCKET_SERVICE` is not set. There are three potential fixes to this issue:

1. Demote from an exception to warning when a local directory from `SIL_NLP_DATA_PATH` is not found.
2. Remove the environment variable `SIL_NLP_DATA_PATH` from each of the clearml agents. 
3. Add a different value for `BUCKET_SERVICE`, such as `any` to each of the clearml agents to force the use of an S3 bucket.

I chose this PR as the best option to not have to restart every agent, and this will allow `SIL_NL_DATA_PATH` to be set and tried first, with S3 as a fallback as long as `BUCKET_SERVICE=""` or is not set.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/678)

### DIFF
--- a/silnlp/common/environment.py
+++ b/silnlp/common/environment.py
@@ -137,9 +137,9 @@ class SilNlpEnv:
                 LOGGER.info(f"Using workspace: {sil_nlp_data_path} as per environment variable SIL_NLP_DATA_PATH.")
                 return Path(sil_nlp_data_path)
             else:
-                raise Exception(
+                LOGGER.warning(
                     f"The path defined by environment variable SIL_NLP_DATA_PATH ({sil_nlp_data_path}) is not a "
-                    + "real directory."
+                    + "local directory."
                 )
 
         gutenberg_path = Path("G:/Shared drives/Gutenberg")


mshannon-sil (commented):
Even if we replace this error with a warning, I think the program would still crash in this scenario. If the datapath is set to /silnlp, it will by default look at MinIO if bucket is "", but the name of our MinIO bucket is nlp-research, not silnlp, so I think it would error out at the end of resolve_data_dir with `raise FileExistsError("No valid path exists")`.

I think we do need to restart the clearml agents with the new SIL_NLP_DATA_PATH pointing to the MinIO bucket, which would mean setting it to /nlp-research right?

I'm also fine with changing this to a warning in addition to changing the ClearML agent, since an empty value for bucket_service can either mean they want to use a local folder, or if not then it defaults to MinIO. Alternatively, we could also add a "local" option for bucket_service and check for "local" here rather than "", which would probably be the cleanest solution so we're not overloading "" but would also be a small breaking change since any users who may have been using a local folder now would need to add this environment variable for that to continue to work.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/678#-:-OLLHH9i0AdbPeYQ7Ep_:b-9tzzvl)* status: 0 of 1 files reviewed, all discussions resolved (waiting on @ddaspit)

TaperChipmunk32 (commented):
The program will not crash in this scenario because of line 151, which sets `sil_nlp_data_path = f"{self.bucket.name}"`. The environment variable is not used for S3 connections, it is only used to check the local file system. In this scenario, this would happen:
1. Check if `\silnlp` is a real directory, if it is not, throw a warning
2. set the s3 bucket to minio
3. set `sil_nlp_data_path` equal to the bucket name
4. set s3_root to `S3Path(sil_nlp_data_path)`
5. return s3_root if it is a directory

> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/678#-:-OLLTTbK0f2Gg0hvLKFR:b-13nhkd)* status: 0 of 1 files reviewed, all discussions resolved (waiting on @ddaspit)

mshannon-sil (approved):
Okay I ran the debugger, and I think what I missed is that the bucket name is hardcoded and not being drawn from SIL_NLP_DATA_PATH. I thought that used to work differently, but I went back and checked and that's what we were doing previously too for the S3 bucket. I'm a little confused then how IDX is using their own S3 bucket to work with SILNLP, but I suppose that's not relevant for this PR. So changing it to a warning should be fine if that will fix it for clearml-session.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/678#-:-OLLbMMY0WCB-03smF8O:b50o4ca)* status: 0 of 1 files reviewed, all discussions resolved (waiting on @ddaspit)

PR #694, opened 2025-03-28
Revised translation options for USFM markers

isaac091:
* Added translation options to toggle placement of paragraph/style/embed markers
* Refactored to move all USFM marker-related work into UsfmPreserver and to always use UsfmPreserver for more consistency across possible configurations

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/694)

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -5174,13 +5174,13 @@ type = ["importlib-metadata (>=7.0.2)", "jaraco.develop (>=7.21)", "mypy (==1.12
 
 [[package]]
 name = "sil-machine"
-version = "1.6.1"
+version = "1.6.2"
 description = "A natural language processing library that is focused on providing tools for resource-poor languages."
 optional = false
 python-versions = "<3.13,>=3.9"
 files = [
-    {file = "sil_machine-1.6.1-py3-none-any.whl", hash = "sha256:38b81f1fc5508e05fe8702c8d0e1a2d5e86e511aa2ce4f30d480744b6a8c8214"},
-    {file = "sil_machine-1.6.1.tar.gz", hash = "sha256:9a4103b5aaaa916f14337b39a1135f5fc40ecff2d46568ed098ac6e99ad07f52"},
+    {file = "sil_machine-1.6.2-py3-none-any.whl", hash = "sha256:89ab7cfe53856c474a1dfb705e90cd3eec6da399b77237005bb16139f43ec353"},
+    {file = "sil_machine-1.6.2.tar.gz", hash = "sha256:25f1818a8a14a4b26e1eacda2f527fbe7ffadba7cdeaa98d450ef5ae6313cc7c"},
 ]
 
 [package.dependencies]
@@ -6217,4 +6217,4 @@ eflomal = ["eflomal"]
 [metadata]
 lock-version = "2.0"
 python-versions = ">=3.10,<3.11"
-content-hash = "1060ce63116411c4c71f57b346a6d2ad59f08f9b2142abc5f000b95b37e14aca"
+content-hash = "8b1efa077c6a39ee15a1f3276f40478d9ce3f1753f956e1d601d7ba6cf6d11de"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -71,7 +71,7 @@ s3path = "0.3.4"
 sacrebleu = "^2.3.1"
 ctranslate2 = "^3.5.1"
 libclang = "14.0.6"
-sil-machine = {extras = ["thot"], version = "^1.6.1"}
+sil-machine = {extras = ["thot"], version = "^1.6.2"}
 datasets = "^2.7.1"
 torch = {version = "^2.4", source = "torch"}
 sacremoses = "^0.0.53"
--- a/silnlp/common/translator.py
+++ b/silnlp/common/translator.py
@@ -17,14 +17,13 @@ from machine.corpora import (
     UpdateUsfmTextBehavior,
     UsfmFileText,
     UsfmStylesheet,
-    UsfmTextType,
     parse_usfm,
 )
 from machine.scripture import VerseRef
 
 from .corpus import load_corpus, write_corpus
 from .paratext import get_book_path, get_iso, get_project_dir
-from .usfm_preservation import NON_NOTE_TYPE_EMBEDS, StatisticalUsfmPreserver
+from .usfm_preservation import StatisticalUsfmPreserver
 
 LOGGER = logging.getLogger(__package__ + ".translate")
 nltk.download("punkt")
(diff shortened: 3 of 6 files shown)


ddaspit (changes requested):
> Reviewed 6 of 6 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/694)* status: all files reviewed, 1 unresolved discussion (waiting on @isaac091)

___
*[`silnlp/nmt/translate.py` line 319 at r1](https://reviewable.io/reviews/sillsdev/silnlp/694#-OMSerodDhtumn-avBGp:-OMSerodDhtumn-avBGq:bnxyo34) ([raw file](https://github.com/sillsdev/silnlp/blob/d95fc3731b4034486f7f062219498b80f241e6e1/silnlp/nmt/translate.py#L319)):*
> ```Python
>     )
>     parser.add_argument(
>         "--include-inline-elements",
> ```

Is it possible to keep the old arguments and translate them to the equivalent new arguments, so that we don't introduce a breaking change. If it doesn't make sense or the EITL team is okay with the breaking change, then we can proceed as is.

isaac091 (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/694)* status: all files reviewed, 1 unresolved discussion (waiting on @ddaspit)

___
*[`silnlp/nmt/translate.py` line 319 at r1](https://reviewable.io/reviews/sillsdev/silnlp/694#-OMSerodDhtumn-avBGp:-OMSh-xV1j5L_cqfwpQK:bczmuvg) ([raw file](https://github.com/sillsdev/silnlp/blob/d95fc3731b4034486f7f062219498b80f241e6e1/silnlp/nmt/translate.py#L319)):*

Is it possible to keep the old arguments and translate them to the equivalent new arguments, so that we don't introduce a breaking change. If it doesn't make sense or the EITL team is okay with the breaking change, then we can proceed as is.
</blockquote></details>

We do need to add at least one additional argument to split "--preserve-usfm-markers" into options for paragraph and style markers separately.
For the embeds, we could potentially keep "--include-inline-elements" since the behavior mostly hasn't changed. The one small difference is that in the current code, if "--preserve-usfm-markers" is used, then the paragraph-type embeds (r/rem) are transferred regardless of the value of "--include-inline_elements". With this update, "--include-embeds" has full control over both paragraph- and character-type embeds. Beyond that, my reasoning for updating the name is (1) we already have confusion with "embeds" vs "notes" and (2) "inline" to me implies "within a verse", i.e. only the character-type embeds. So, I would vote to keep the name change as a clean break from the already reused argument, but if it's going to cause issues I can switch back.

ddaspit (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/694)* status: all files reviewed, 1 unresolved discussion (waiting on @isaac091)

___
*[`silnlp/nmt/translate.py` line 319 at r1](https://reviewable.io/reviews/sillsdev/silnlp/694#-OMSerodDhtumn-avBGp:-OMT14iZ7IIwn4r64fee:bcmpbqq) ([raw file](https://github.com/sillsdev/silnlp/blob/d95fc3731b4034486f7f062219498b80f241e6e1/silnlp/nmt/translate.py#L319)):*
<details><summary><i>Previously, isaac091 (Isaac Schifferer) wrote…</i></summary><blockquote>

We do need to add at least one additional argument to split "--preserve-usfm-markers" into options for paragraph and style markers separately.
For the embeds, we could potentially keep "--include-inline-elements" since the behavior mostly hasn't changed. The one small difference is that in the current code, if "--preserve-usfm-markers" is used, then the paragraph-type embeds (r/rem) are transferred regardless of the value of "--include-inline_elements". With this update, "--include-embeds" has full control over both paragraph- and character-type embeds. Beyond that, my reasoning for updating the name is (1) we already have confusion with "embeds" vs "notes" and (2) "inline" to me implies "within a verse", i.e. only the character-type embeds. So, I would vote to keep the name change as a clean break from the already reused argument, but if it's going to cause issues I can switch back.
</blockquote></details>

I like the new arguments. I don't think we should remove those. I was just thinking that it might be good to keep the old arguments if we can easily translate them to the new arguments. For example, `--include-inline-elements` might set `include_embeds` to `True`. I'm not sure what `--preserve-usfm-markers` does, but maybe that is the equivalent to setting `include_paragraph_markers` and `include_style_markers` to `True`. Does that make sense?

isaac091 (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/694)* status: all files reviewed, 1 unresolved discussion (waiting on @ddaspit)

___
*[`silnlp/nmt/translate.py` line 319 at r1](https://reviewable.io/reviews/sillsdev/silnlp/694#-OMSerodDhtumn-avBGp:-OMT48qy8gwpb_qmvjpZ:b-yngdgl) ([raw file](https://github.com/sillsdev/silnlp/blob/d95fc3731b4034486f7f062219498b80f241e6e1/silnlp/nmt/translate.py#L319)):*

I like the new arguments. I don't think we should remove those. I was just thinking that it might be good to keep the old arguments if we can easily translate them to the new arguments. For example, `--include-inline-elements` might set `include_embeds` to `True`. I'm not sure what `--preserve-usfm-markers` does, but maybe that is the equivalent to setting `include_paragraph_markers` and `include_style_markers` to `True`. Does that make sense?
</blockquote></details>

Ok, I see. Yes, I can do that.

ddaspit (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/694)* status: all files reviewed, 1 unresolved discussion (waiting on @isaac091)

___
*[`silnlp/nmt/translate.py` line 319 at r1](https://reviewable.io/reviews/sillsdev/silnlp/694#-OMSerodDhtumn-avBGp:-OMT4i-7DbM3ST1r037f:b-1305t0) ([raw file](https://github.com/sillsdev/silnlp/blob/d95fc3731b4034486f7f062219498b80f241e6e1/silnlp/nmt/translate.py#L319)):*
<details><summary><i>Previously, isaac091 (Isaac Schifferer) wrote…</i></summary><blockquote>

Ok, I see. Yes, I can do that.
</blockquote></details>

In the `help` property, you should indicate that the old arguments are deprecated.

isaac091 (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/694)* status: 4 of 6 files reviewed, 1 unresolved discussion (waiting on @ddaspit)

___
*[`silnlp/nmt/translate.py` line 319 at r1](https://reviewable.io/reviews/sillsdev/silnlp/694#-OMSerodDhtumn-avBGp:-OMTCWPa9SKGZEqg3Yjn:b-896fix) ([raw file](https://github.com/sillsdev/silnlp/blob/d95fc3731b4034486f7f062219498b80f241e6e1/silnlp/nmt/translate.py#L319)):*

In the `help` property, you should indicate that the old arguments are deprecated.
</blockquote></details>

Done.

ddaspit (approved):
> Reviewed 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/694)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @isaac091)

PR #700, opened 2025-04-05
Combine scores

davidbaines:
Fix a bug that was preventing the correct path from being found.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/700)


davidbaines:
I'm updating combine_scores branch from the master branch. No need to change master.
At the moment I'm using the combine_scores since so much on master requires eflomal, and eflomal can't be installed on Windows.

PR #709, opened 2025-04-16
Move the analyze_project_pairs.py from nmt to common folder.  Rename it to align.py  Add to the output information.

davidbaines:
I found that I make many mistakes when I'm trying to use the analyze_project_pairs script.

I think it should be in common since it isn't using NMT, so I often type   `python -m silnlp.common.analyze.project_pairs`
For some reason I expect it to be analyze_corpus_pairs, so I'm often typing that too.  Finally I often use the English spelling and put analyse_project_pairs.  

Moving this to the common folder, where it used to be, and renaming it to align.py will solve all these issues.
I've also added more information to the output that is logged, that will make it easier to see which alignment is currently being calculated, which helps to keep track of progress.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/709)


isaac091 (commented):
It makes sense to move the script to `common`, but I think `align.py` might also be confusing because 1. it's not only doing alignment, though that is the main function of it, and 2. we already have an `alignment/align.py`. What about `get_corpus_stats.py` or `compute_corpus_stats.py`?

> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/709#-:-ONzmKtL7z9AGrs4stse:b-2mm5rv)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @davidbaines)

davidbaines:
Thanks for your review Isaac!
I saw that there was an align.py later, but was really hoping that wouldn't be a problem since they do different things and are in different folders.

I'm keen to keep the name short, since the name is essentially part of the UI as far as the EITL team are concerned. We have to remember it and type it. 
If we really have to rename it, how about naming this alignment.py?

isaac091 (approved):
It makes sense that you want to keep the name short, and since alignment.py is not that different, I'm OK to stick with align.py.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/709#-:-OOTQ1FXBxrGKJTNSg4X:bcpjshd)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @davidbaines)

Issue #715, opened 2025-04-24
Add confidence metrics to test.py, diff_predictions.py, translate.py


mshannon-sil:
Some of our research for Catapult Reloaded/Crossbow will involve looking at confidence scores to help guide our understanding of chapter quality. We need test.py to output the confidence score for each sequence/verse along with individual token confidences, and then diff_predictions.py can use these verse level confidence scores to compute a chapter level confidence score.

mshannon-sil:
In addition, we need the ability to output confidence scores by book through the --score-by-book feature, so that we can see if the correlation between chrF3 and confidence holds across books. And we need to add confidence scores to the translate step, since this is what would be used in a real scenario to obtain confidences for a draft rather than the test step.

PR #716, opened 2025-04-24
Add confidence scores as output for test.py, diff_predictions.py, translate.py

mshannon-sil:
This addresses issue #715. The `test.py` script now outputs a file called `test.trg-predictions.txt.[ckpt_num].confidences.tsv`. The file has two rows for each verse. The first row has the verse number and then each token, and the second row has the sequence confidence and then each individual token confidence, so that they are vertically paired with the first row. The `diff_predictions.py` script now has a couple new flags, `--chapter-score` which @AmeWenJ added that runs diff predictions at a chapter level rather than a verse level, and `--confidence` which adds a column to the dataframe for confidence scores.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/716)

### DIFF
--- a/silnlp/common/environment.py
+++ b/silnlp/common/environment.py
@@ -252,7 +252,9 @@ class SilNlpEnv:
                     LOGGER.info("Downloading " + rel_path)
                     try_n_times(lambda: self.bucket.download_file(obj.object_key, str(temp_dest_path)))
 
-    def copy_experiment_from_bucket(self, name: Union[str, Path], patterns: Union[str, Sequence[str]] = []):
+    def copy_experiment_from_bucket(
+        self, name: Union[str, Path], patterns: Union[str, Sequence[str]] = [], no_checkpoints: bool = False
+    ):
         if not self.is_bucket:
             return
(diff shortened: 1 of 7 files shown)


mshannon-sil:
I realize that I should also add this to the translate step, not just the test step, since that's what would be used in a real project scenario. I'll work on adding that and include it as part of this PR.

ddaspit (approved):
It's always fun when you get to do PyTorch programming. Good job.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/716#-:-OOilzn-BeHzvPUcCfQB:b-oxq3g7)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @AmeWenJ and @benjaminking)

AmeWenJ (approved):
Thank you for merging with new functionality!

AmeWenJ (approved):

AmeWenJ (approved):

AmeWenJ (approved):

PR #718, opened 2025-04-30
Revert to older implementation of USFM marker placement

isaac091:
This is the implementation of marker placement pre- any related updates to machine.py, but keeping as many improvements made to this code since then. The only obvious way that output USFM will be different is that all of the paragraph markers for a verse will be present, even when they are not being placed, as it was before any marker placement-related updates.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/718)

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -5174,13 +5174,13 @@ type = ["importlib-metadata (>=7.0.2)", "jaraco.develop (>=7.21)", "mypy (==1.12
 
 [[package]]
 name = "sil-machine"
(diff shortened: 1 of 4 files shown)


ddaspit (approved):
> Reviewed 4 of 4 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/718)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @isaac091)

Issue #725, opened 2025-05-10
Automated scheduled cleanup of Paratext projects folder


mmartin9684-sil:
Could we run a script over the Paratext > projects folder on a regular basis to remove PII from the projects, as well as other project data that is not needed by the silnlp tools (graphics, .hg folder, etc)?
Automatically removing PII would help us ensure we more consistently follow good security practices for all projects.  Automatically removing unnecessary project artifacts, such as graphics and revision history, will reduce storage demands.
David has a script that he runs on occasion that could be a starting point.

PR #729, opened 2025-05-22
Replace clean_projects with new script

davidbaines:
A new script for removing unnecessary data from project folders. 
Possibly this could be part of the solution to issue #725 
Other wise it is a script I can run occasionally manually. 

I don't think that it removes any data that we make use of, but it may well be removing data that we might possibly make use of in the future. I'm trying to err on the side of caution regarding the removal of PII from our shared project folders.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/729)

### DIFF
--- a/silnlp/common/clean_projects.py
+++ b/silnlp/common/clean_projects.py
@@ -1,0 +1,441 @@
+#!/usr/bin/env python3
+
+import argparse
+import logging
+import shutil
+import sys
+from pathlib import Path
+from typing import Optional
+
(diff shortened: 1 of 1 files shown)


benjaminking (approved):
Nice work. :lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/729#-:-OQx6NSJBjKDp8a9mOIk:bb4nfal)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @mshannon-sil)

Issue #746, opened 2025-06-03
Generating multiple drafts of a book with different draft formatting settings in one experiment


mmartin9684-sil:
With several new settings for controlling the formatting of drafts (paragraph marker placement, quotation denormalization, style marker placement), it would be convenient to be able to request drafts of the same book with different settings for these options.  This would primarily be for internal testing of these new features, but also for purposes of introducing new projects to the impacts of these settings.

Currently, in order to create multiple drafts for the same book with different format settings, its necessary to make multiple runs of the experiment command.  The first run would be for the full experiment (preprocess/train/test/translate) with one drafting configuration, followed by changes to the translate_config.yml, and then a second (partial) run of the experiment script (preprocess/translate or test/translate).   In between the first and second runs of the experiment script, manual updates are need to the `infer > checkpoint > source` folder names, so that the first draft isn't overwritten by the second draft.

A single run of the experiment script that creates multiple drafts of the same book with varying drafting configurations would be simpler, less error prone, and less time-consuming.

isaac091:
The first simple step towards something like this would be to save a copy of the draft with no post-processing done to it whenever some post-processing steps have been configured for. 

Beyond that, we can have a discussion about what additional flexibility would be useful. The ability to have multiple translation configurations while only translating the text once would be new to silnlp, and so would take more work to implement, though it will probably be straightforward once we decide on a design that is intuitive for you all as users.

@davidbaines had also asked about being able to apply different post-processing after the fact. This would be relatively simple to implement as well, probably in a separate script.

I can knock out those couple of things and then we can discuss further as needed.

isaac091:
Now that I think about it, a draft with no post-processing is not actually needed to produce another draft with alternate post-processing because, at least in the realm of USFM markers, the markers can be stripped out and placed back in again. So I'll hold off on making the translate step save a base draft, but once we have the more detailed configuration there will be a way to have one of the drafts be a base draft.

isaac091:
@mmartin9684-sil @davidbaines 

I added the script `silnlp.common.postprocess_draft` that isolates the marker placement process (and in the future other features under the 'Improving Draft Quality' umbrella) from the rest of the translate pipeline. The corresponding section of the wiki is [here](https://github.com/sillsdev/silnlp/wiki/NMT:-Usage#postprocess_draft).

[...]

PR #750, opened 2025-06-05
Add script for creating drafts with alternate postprocessing configurations

isaac091:
Towards #746

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/750)

### DIFF
--- a/silnlp/common/postprocess_draft.py
+++ b/silnlp/common/postprocess_draft.py
@@ -1,0 +1,163 @@
+import argparse
+import logging
+import re
+from pathlib import Path
+from typing import List, Tuple
+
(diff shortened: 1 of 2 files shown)


isaac091:
The intention is that this script is only used with unedited drafts (i.e. straight from the translate pipeline) and their original sources. This is because the ScriptureRefs of the source and draft have to match up exactly (with the exception of \rem elements) so that alignment can be done for marker placement. So at least in the way that I'm thinking about this script right now, it would have access to the quotation convention of the source and assume that the quotations in the draft are either normalized or have the same convention as the source.

But yes, we will need to have that conversation.

PR #753, opened 2025-06-11
Update lex tools

bhartmoore:
Add functionality to support "Novel words" investigations for Catapult Reloaded and Crossbow work. Updates include: 
- Order source/target words by occurrences in given book(s) and print a count of occurrences
- Output a list of verses where "novel" words occur, as well as the words that occur in those verses with their counts

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/753)


mshannon-sil (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/753)* status: all files reviewed, 5 unresolved discussions (waiting on @bhartmoore)

___
*[`silnlp/common/find_words.py` line 32 at r1](https://reviewable.io/reviews/sillsdev/silnlp/753#-OSVj3Su69etiPSiJjU6:-OSVj3Su69etiPSiJjU7:b8rz64b) ([raw file](https://github.com/sillsdev/silnlp/blob/d68f27a9d51ddfe76eea8f4a470561930b0ea3e2/silnlp/common/find_words.py#L32)):*
> ```Python
>                     src_word_counts.append(entry)
>             else:
>                 print("Error: word counts are missing. Please run count_words.py with the --count flag set.")
> ```

I think it would be more appropriate to raise an exception here, unless there's a strong reason you need it to complete successfully. You could do `raise ValueError("Word counts are missing. Please run count_words.py with the --count flag set.")`.
___
*[`silnlp/common/find_words.py` line 44 at r1](https://reviewable.io/reviews/sillsdev/silnlp/753#-OSVnKu78tXAON-5fYku:-OSVnKu88XeTkD5BUC1T:b1otq0b) ([raw file](https://github.com/sillsdev/silnlp/blob/d68f27a9d51ddfe76eea8f4a470561930b0ea3e2/silnlp/common/find_words.py#L44)):*
> ```Python
>     with (lex_path / "src.txt").open("r", encoding = "utf8") as src_data_file:
>             with(vref_filename).open("r", encoding = "utf8") as ref_file:
>                 word_list = list(enumerate(words))
> ```

It doesn't seem that word_list[0] is used in any of the following code, only word_list[1], so I don't think you need to use `enumerate`, unless it's there for debugging purposes.
___
*[`silnlp/common/find_words.py` line 61 at r1](https://reviewable.io/reviews/sillsdev/silnlp/753#-OSVnuXQ63sZx_9c5AM8:-OSVnuXQ63sZx_9c5AM9:bfyf54a) ([raw file](https://github.com/sillsdev/silnlp/blob/d68f27a9d51ddfe76eea8f4a470561930b0ea3e2/silnlp/common/find_words.py#L61)):*
> ```Python
>                             word_text.append(word[1])
>                             word_num.append(src_word_dict[word[1]])
>                     result.append([verse[0].rstrip('\n'), word_count, word_num, word_text])
> ```

What does `word_count` here mean? It seems that it's the sum of the counts for all the words that are in the word list, appear in this verse, and haven't been seen yet. But I'm not sure what this metric means or is used for, so I want to check if my understanding is correct and this is the desired functionality.
___
*[`silnlp/common/compare_lex.py` line 10 at r1](https://reviewable.io/reviews/sillsdev/silnlp/753#-OSVgERGDE6_OddfPbnM:-OSVgERGDE6_OddfPbnN:b7t9mwk) ([raw file](https://github.com/sillsdev/silnlp/blob/d68f27a9d51ddfe76eea8f4a470561930b0ea3e2/silnlp/common/compare_lex.py#L10)):*
> ```Python
> from machine.tokenization import LatinWordTokenizer
> 
> # Latin Tokenizer from machine library
> ```

Are these comments still needed?
___
*[`silnlp/common/compare_lex.py` line 100 at r1](https://reviewable.io/reviews/sillsdev/silnlp/753#-OSVgeI60Cs_0M1MAuS9:-OSVgeI60Cs_0M1MAuSA:b-dxpjc9) ([raw file](https://github.com/sillsdev/silnlp/blob/d68f27a9d51ddfe76eea8f4a470561930b0ea3e2/silnlp/common/compare_lex.py#L100)):*
> ```Python
> 
>         # Write source words missing from the alternate source file
>         #with (lex_path1 / "unmatched_src_words.txt").open("w", encoding="utf8") as output_file:
> ```

Are these comments also still needed?

mshannon-sil:
We discussed that maybe in the future we will want to research more regarding novel vocab and flesh out the implementation, but for now it's not planned and we can close this out.

PR #755, opened 2025-06-13
Add ability to run postprocessing with an experiment and with ClearML

isaac091:
I realized that Windows users won't be able to use this without running on ClearML because of eflomal, so unfortunately this script isn't as convenient for the EITL team as I had originally thought. Regardless, I think it will be useful in the future once eflomal  is usable outside of Linux and also more generally if/when we add more post-processing options.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/755)

### DIFF
--- a/silnlp/common/postprocess_draft.py
+++ b/silnlp/common/postprocess_draft.py
@@ -4,6 +4,7 @@ import re
 from pathlib import Path
 from typing import List, Tuple
 
+import yaml
 from machine.corpora import (
     FileParatextProjectSettingsParser,
     ScriptureRef,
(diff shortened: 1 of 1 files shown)


benjaminking (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/755)* status: all files reviewed, 1 unresolved discussion (waiting on @isaac091)

___
*[`silnlp/common/postprocess_draft.py` line 142 at r1](https://reviewable.io/reviews/sillsdev/silnlp/755#-OSt9Ixs-tJ5joe_5BaI:-OSt9Ixs-tJ5joe_5BaJ:b-kq57f1) ([raw file](https://github.com/sillsdev/silnlp/blob/836f733226b6f148056af28f244a1ca1885ac4bd/silnlp/common/postprocess_draft.py#L142)):*
> ```Python
> 
>     source_or_experiment = args.source_or_experiment.replace("\\", "/")
>     if get_mt_exp_dir(source_or_experiment).exists():
> ```

I think we should restructure the arguments so that the user explicitly specifies whether they're providing an experiment or a file.  Just thinking of cases like where you're running it on a file named "test" and there's also an experiment named "test"

benjaminking (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/755#-:-OSuLaOaAy_fE7PZx7KU:bnfp4nl)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @isaac091)

PR #757, opened 2025-06-19
Ignore remarks and empty end-of-verse paragraphs

isaac091:
Since the draft remark only gets added to the draft, the vrefs were not aligned between the two files.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/757)

### DIFF
--- a/silnlp/common/compare_usfm_structure.py
+++ b/silnlp/common/compare_usfm_structure.py
@@ -100,15 +100,19 @@ def evaluate_usfm_marker_placement(
 
     tokenizer = WhitespaceMarkerTokenizer()
 
+    gold_file_sents = [
+        gs.text for gs in gold_file_text if not (len(gs.ref.path) > 0 and gs.ref.path[-1].name in PARAGRAPH_TYPE_EMBEDS)
+    ]
+    pred_file_sents = [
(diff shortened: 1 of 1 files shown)


benjaminking (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/757)* status: all files reviewed, 1 unresolved discussion (waiting on @isaac091)

___
*[`silnlp/common/compare_usfm_structure.py` line 103 at r1](https://reviewable.io/reviews/sillsdev/silnlp/757#-OT8PNGGAhLqs--V-t_u:-OT8PNGH967Z7jLcRX-O:b-wuu186) ([raw file](https://github.com/sillsdev/silnlp/blob/97d7781f63201fbf99efefc02ac380b8c793ff6c/silnlp/common/compare_usfm_structure.py#L103)):*
> ```Python
>     tokenizer = WhitespaceMarkerTokenizer()
> 
>     gold_file_sents = [
> ```

It might be helpful to either add a comment or split this out into a separate method with a descriptive name, since what it's doing is a little unintuitive.

isaac091:
I added another commit that is completely different from the original one but that is doing a similar type of thing as far as excluding elements that we don't want to factor into the score. A recent update to the USFM parser has it include the empty end-of-verse paragraph markers in row outputs, and that was artificially inflating the similarity scores.

PR #759, opened 2025-06-20
Add support for multiple postprocessing requests

### LINKED ISSUE #746: Generating multiple drafts of a book with different draft formatting settings in one experiment

isaac091:
This reworks how translate configs are expected to look. Instead of specifying the postprocessing options inside of each translate request, there will be a separate 'postprocess' section at the top level which is a list of postprocessing requests that will be applied to all translate requests. A draft with no postprocessing applied is always saved. Once this is pushed, I will update the [wiki](https://github.com/sillsdev/silnlp/wiki/Configure-a-model#how-to-configure-translation-requests-for-a-model) with an example and explanation.

Example:
```
translate:
  - src_project: NASB
    books: MAT
  - src_project: NIV11R
    books: MAT
postprocess:
  - include_paragraph_markers: True  # only paragraph markers
  - include_paragraph_markers: True  # all markers
    include_style_markers: True
    include_embeds: True
```

It also changes the behavior of `postprocess_draft.py` slightly. Now, if none of the postprocessing options are toggled and the `--experiment` option is used, the script will apply any postprocessing requests in the experiment's translate config. Previously, if no postprocessing options were used, it would create a base draft, but that is no longer necessary because they are created by default, and I can foresee folks expecting to be able to use the translate config to configure this script. I will also update the wiki to reflect these changes.

This PR is ready to be reviewed, but not to push right away. I still need to get feedback on what the names of the output files should look like. Currently, the base draft will have the same file name as drafts do now, and any outputs with marker placement options will have a suffix, e.g. `41MAT_pse.SFM`, that indicates if `p`aragraph markers, `s`tyle markers, or `e`mbeds were included.

Closes #746

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/759)


isaac091 (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/759)* status: 4 of 9 files reviewed, 4 unresolved discussions (waiting on @benjaminking)

___
*[`silnlp/common/postprocesser.py` line 58 at r4](https://reviewable.io/reviews/sillsdev/silnlp/759#-OU6St3uDqE4ZoMtAzeq:-OUGneMf2rvtnYYgZH1E:b-896fix) ([raw file](https://github.com/sillsdev/silnlp/blob/ff04368795e9a280ac19467187b15026932b69a9/silnlp/common/postprocesser.py#L58)):*

I think it might be better if this class could be decoupled from the way that postprocess configs are specified in `translate_config.yml`. Specifically, I'm thinking of having the constructor take a list of `PostprocessConfig`s.
</blockquote></details>

Done.

benjaminking (approved):
:lgtm:

> Reviewed 5 of 5 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/759#-:-OU_2EK92t2zbEfq1cal:bnfp4nl)* status: all files reviewed, 4 unresolved discussions (waiting on @isaac091)

PR #767, opened 2025-07-07
Postprocess backwards compatibility

isaac091:
Checks for and uses post-processing options defined in translate requests if no 'postprocess' section exists in the translate config. I meant to do this in the last PR but forgot to.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/767)


benjaminking (changes requested):
> Reviewed 5 of 5 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/767)* status: all files reviewed, 3 unresolved discussions (waiting on @isaac091)

___
*[`silnlp/common/postprocess_draft.py` line 102 at r1](https://reviewable.io/reviews/sillsdev/silnlp/767#-OUaf1P69d70T3cqVXm4:-OUaf1P69d70T3cqVXm5:b2webjw) ([raw file](https://github.com/sillsdev/silnlp/blob/8afb1885c63e116ded29f90c493c77bc07ab1e8f/silnlp/common/postprocess_draft.py#L102)):*
> ```Python
>             LOGGER.info("No postprocessing options used. Applying postprocessing requests from translate config.")
>             postprocess_experiment(config, args.output_folder)
>             exit()
> ```

I think you could avoid needing to use an "exit" if you put the last few lines of this file under the "else."
___
*[`silnlp/common/postprocess_draft.py` line 114 at r1](https://reviewable.io/reviews/sillsdev/silnlp/767#-OUafGVcD3UN0Ro2O7Nk:-OUafGVdCCFP2DjHaev1:bddg8ul) ([raw file](https://github.com/sillsdev/silnlp/blob/8afb1885c63e116ded29f90c493c77bc07ab1e8f/silnlp/common/postprocess_draft.py#L114)):*
> ```Python
> 
>     if len(postprocess_config.keys()) == 0:
>         LOGGER.info("Please use at least one postprocessing option.")
> ```

This would probably be better as an exception, since it causes execution to terminate.
___
*[`silnlp/nmt/postprocess.py` line 187 at r1](https://reviewable.io/reviews/sillsdev/silnlp/767#-OUainIh69BIPTJbFt99:-OUainIh69BIPTJbFt9A:b-xielkp) ([raw file](https://github.com/sillsdev/silnlp/blob/8afb1885c63e116ded29f90c493c77bc07ab1e8f/silnlp/nmt/postprocess.py#L187)):*
> ```Python
> 
> def main() -> None:
>     parser = argparse.ArgumentParser(description="Postprocess the drafts created by an NMT model")
> ```

Could you include something that explains the difference between postprocess.py and postprocess_draft.py in the descriptions of the two scripts?  After not looking at them for a few days, I was having trouble remembering which one did which thing.

PR #768, opened 2025-07-08
Guess book id when translating isolated USFM files

isaac091:
This was a leftover issue from when we switched to using the machine USFM parser. I say it's a guess because we're not extracting the book ID from the file name and I don't know if there's an explicit requirement for the \id line to start with the 3-character book code, but I think in reality they all do.

This also fixes an issue I ran into while testing, the "patterns" arguments are just left over from the days of non-mounted buckets.

### DIFF
--- a/silnlp/common/translator.py
+++ b/silnlp/common/translator.py
@@ -178,7 +178,11 @@ class Translator(ABC):
                 project=src_settings.name,
             )
         else:
-            src_file_text = UsfmFileText("usfm.sty", "utf-8-sig", "", src_file_path, include_all_text=True)
+            # Guess book ID
+            with src_file_path.open(encoding="utf-8-sig") as f:
+                book_id = f.read().split()[1].upper()
+
+            src_file_text = UsfmFileText("usfm.sty", "utf-8-sig", book_id, src_file_path, include_all_text=True)
         stylesheet = src_settings.stylesheet if src_from_project else UsfmStylesheet("usfm.sty")
 
         sentences = [re.sub(" +", " ", s.text.strip()) for s in src_file_text]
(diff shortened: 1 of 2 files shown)


benjaminking (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/768)* status: all files reviewed, 1 unresolved discussion (waiting on @isaac091)

___
*[`silnlp/common/translator.py` line 183 at r1](https://reviewable.io/reviews/sillsdev/silnlp/768#-OUf_UskD9VBsgkl0HsY:-OUf_UskD9VBsgkl0HsZ:b861fsz) ([raw file](https://github.com/sillsdev/silnlp/blob/af0a8b01298ca462a3de62d6baaf30f4258356d3/silnlp/common/translator.py#L183)):*
> ```Python
>             # Guess book ID
>             with src_file_path.open(encoding="utf-8-sig") as f:
>                 book_id = f.read().split()[1].upper()
> ```

It would probably be worth adding a sanity check that the second token here looks like a book ID (i.e. 3 alphanumeric characters)

isaac091 (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/768)* status: 1 of 2 files reviewed, 1 unresolved discussion (waiting on @benjaminking)

___
*[`silnlp/common/translator.py` line 183 at r1](https://reviewable.io/reviews/sillsdev/silnlp/768#-OUf_UskD9VBsgkl0HsY:-OUffmqb9uzyQpQjj1Kc:b-896fix) ([raw file](https://github.com/sillsdev/silnlp/blob/af0a8b01298ca462a3de62d6baaf30f4258356d3/silnlp/common/translator.py#L183)):*

It would probably be worth adding a sanity check that the second token here looks like a book ID (i.e. 3 alphanumeric characters)
</blockquote></details>

Done.

PR #783, opened 2025-08-01
Improved onboard_project.py

TaperChipmunk32:
-Added collect_verse_counts to onboard_project
-Replaced command line args with a yaml config

### DIFF
--- a/silnlp/common/collect_verse_counts.py
+++ b/silnlp/common/collect_verse_counts.py
@@ -264,10 +264,8 @@ def main() -> None:
     # If the output folder doesn't exist locally, assume it's an experiment folder
     folder = args.folder.replace("\\", "/")
     if Path(folder).exists():
-        exp_name = None
         folder = Path(folder)
     else:
-        exp_name = folder
         folder = get_mt_exp_dir(folder)
 
     # If no files are listed and folder is an experiment, use the files listed in the config file
--- a/silnlp/common/extract_corpora.py
+++ b/silnlp/common/extract_corpora.py
@@ -105,6 +105,7 @@ def extract_corpora(
                 extract_lemmas,
                 extract_project_vrefs,
             )
+            LOGGER.info(f"Extracted corpus file: {corpus_filename}")
             # check if the number of lines in the file is correct (the same as vref.txt)
             LOGGER.info(f"# of Verses: {verse_count}")
             if verse_count != expected_verse_count:
--- a/silnlp/common/onboard_project.py
+++ b/silnlp/common/onboard_project.py
@@ -2,6 +2,8 @@ import argparse
 import logging
 from pathlib import Path
 
+import yaml
+
 from .environment import SIL_NLP_ENV
 
 LOGGER = logging.getLogger(__package__ + ".onboard_project")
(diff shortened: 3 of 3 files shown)


TaperChipmunk32 (commented):
I've now added some messages at a high level, extract_corpora and collect_verse_counts will both print more messages as they execute.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/783#-:-OWbGYg79K2tcIpSpjWO:b-z3e52k)* status: 2 of 3 files reviewed, 3 unresolved discussions (waiting on @benjaminking and @ddaspit)

___
*[`silnlp/common/onboard_project.py` line 92 at r1](https://reviewable.io/reviews/sillsdev/silnlp/783#-OWascCI9yxYu0ok_53b:-OWbGqeqDLsevLCAtzm1:b-w5m46r) ([raw file](https://github.com/sillsdev/silnlp/blob/e646dff9c632b06092f4f4e5d5f9bd15c9bd933a/silnlp/common/onboard_project.py#L92)):*

I don't think you should need to parse the arguments again.
</blockquote></details>

I don't know how I did not catch these repeated lines, I've now removed all the copied lines.
___
*[`silnlp/common/onboard_project.py` line 98 at r1](https://reviewable.io/reviews/sillsdev/silnlp/783#-OWat3xDBBhMbiH7lEK1:-OWbH62TEKvGCJ7XtfL0:b-ywyjc4) ([raw file](https://github.com/sillsdev/silnlp/blob/e646dff9c632b06092f4f4e5d5f9bd15c9bd933a/silnlp/common/onboard_project.py#L98)):*

Maybe we should add an error if the user doesn't include either of `copy_from` or `config`, since the script will say it's onboarding the project, but then only create a new folder at the most.
</blockquote></details>

I've now added some error messages if they are not included.
___
*[`silnlp/common/onboard_project.py` line 106 at r1](https://reviewable.io/reviews/sillsdev/silnlp/783#-OWatmXP8ATHlaFqLgt_:-OWbHE4u2NH04mLh0_2a:b-q613uu) ([raw file](https://github.com/sillsdev/silnlp/blob/e646dff9c632b06092f4f4e5d5f9bd15c9bd933a/silnlp/common/onboard_project.py#L106)):*

Now that we have a config file, I think we should start a page on the wiki ([Home · sillsdev/silnlp Wiki](https://github.com/sillsdev/silnlp/wiki/)) that we keep up to date as we add new configuration options.
</blockquote></details>

I'll start working on that now.

benjaminking (changes requested):
You'll probably need to have Damien take over reviewing this while I'm gone.  We're pretty close here though.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/783#-:-OWjiMyt91L9RDSdwBan:bp5vnvk)* status: all files reviewed, 4 unresolved discussions (waiting on @ddaspit)

___
*[`silnlp/common/onboard_project.py` line 123 at r2](https://reviewable.io/reviews/sillsdev/silnlp/783#-OWjgCm2-ghMkW9u53a9:-OWjgCm2-ghMkW9u53aA:bpm645g) ([raw file](https://github.com/sillsdev/silnlp/blob/ce3021138c33ba51ed0a77a9015bf892bbd644c7/silnlp/common/onboard_project.py#L123)):*
> ```Python
>         config = {}
> 
>     if args.extract_corpora:
> ```

I expect that the EITL team will want to have a standard config for onboarding projects and will just re-use it for each new project they onboard.  What do you think about keeping `extract_corpora` and `select_verse_counts` in the config, but allowing the command line options to override what's in the config?
___
*[`silnlp/common/onboard_project.py` line 139 at r2](https://reviewable.io/reviews/sillsdev/silnlp/783#-OWjh4jN0bphOWg6u_Ot:-OWjh4jN0bphOWg6u_Ou:b-6f2ukx) ([raw file](https://github.com/sillsdev/silnlp/blob/ce3021138c33ba51ed0a77a9015bf892bbd644c7/silnlp/common/onboard_project.py#L139)):*
> ```Python
> 
>     if args.collect_verse_counts:
>         if not args.extract_corpora:
> ```

Let me know if what I'm thinking isn't feasible, but could we check whether the files that we need for `collect_verse_counts` are present already?  The scenario that I'm imagining is that someone runs this script, but later realizes that they forgot to collect verse counts.  Would it be possible for them to run it again with only `collect_verse_counts` enabled (and `overwrite` equal to false)?

davidbaines:
@benjaminking  suggested: I expect that the EITL team will want to have a standard config for onboarding projects and will just re-use it for each new project they onboard. What do you think about keeping extract_corpora and select_verse_counts in the config, but allowing the command line options to override what's in the config?

That sounds like a great idea!

If there's one onboarding config for most projects the name is less important. If there's one per project then we could name it onboarding_config.yml to distinguish it from the experiment configs.  Hopefully we can use a small number of onboarding configs that aren't stored in the experiment folder and we choose one that is appropriate for each project.

benjaminking (approved):
:lgtm:

> @benjaminking reviewed 1 of 1 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/783#-:-OYI5bH_6iyMleG-SGp5:bnfp4nl)* status: all files reviewed, 2 unresolved discussions (waiting on @ddaspit and @TaperChipmunk32)

PR #793, opened 2025-08-13
Add quality estimation module

mshannon-sil:
This module adds support for estimating the quality of drafts with regard to confidence, chrf3, and usability.

It requires a diff_predictions file to establish a correlation between confidence and chrF3, as well as confidence files to project chrF3 and ultimately usability from. The user has a few options to specify confidence files. They can provide a list of file paths (relative to the experiment directory) or they can specify just the book ids. They can also change the directory to look for confidence files in (relative to the experiment directory) e.g. /infer/5000/source. See the argument descriptions for more details. Also, if a usability_parameters.tsv file is present, it will use the parameters in that file for the chrf3 - usability distribution rather than the default parameters that we gathered from usability surveys.

It outputs '*.projected_chrf3.tsv' files corresponding to each confidence file, listing the vref, confidence, and projected_chrf3 for each verse. It then also outputs a single usability_chapters.tsv file covering chapter usability for all chapter refs in all the confidence files, as well as a single usability_books.tsv for book usability.

Currently it only supports chrF3 as a projected scorer and assumes confidence files were generated from a .SFM file rather than a .txt file so that vref information can be retrieved. Future work could be done to expand functionality as the need arises.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/793)

### DIFF
--- a/silnlp/nmt/quality_estimation.py
+++ b/silnlp/nmt/quality_estimation.py
@@ -23,14 +23,12 @@ class VerseScore:
     projected_chrf3: Optional[float] = None
 
 
-def estimate_quality(diff_predictions_file: Path, confidence_files: List[Path]):
-    verse_scores: List[VerseScore] = []
-    for confidence_file in confidence_files:
-        verse_scores += project_chrf3(diff_predictions_file, confidence_file)
+def estimate_quality(diff_predictions_file: Path, confidence_files: List[Path]) -> None:
+    verse_scores: List[VerseScore] = project_chrf3(diff_predictions_file, confidence_files)
     compute_usable_proportions(verse_scores, confidence_files[0].parent)
 
 
-def project_chrf3(diff_predictions_file: Path, confidence_file: Path) -> List[VerseScore]:
+def project_chrf3(diff_predictions_file: Path, confidence_files: List[Path]) -> List[VerseScore]:
     chrf3_scores, confidence_scores = extract_diff_predictions(diff_predictions_file)
     if len(chrf3_scores) != len(confidence_scores):
         raise ValueError(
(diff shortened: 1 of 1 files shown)


mshannon-sil (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/793)* status: 0 of 1 files reviewed, 6 unresolved discussions (waiting on @benjaminking)

___
*[`silnlp/nmt/quality_estimation.py` line 26 at r3](https://reviewable.io/reviews/sillsdev/silnlp/793#-OXyXzMo7FpRcB_LEfqf:-OYDLvhGBTd0PiW72sYl:b-896fix) ([raw file](https://github.com/sillsdev/silnlp/blob/4efd39b802831aeb00c07313899054c2c79bcd04/silnlp/nmt/quality_estimation.py#L26)):*

Can you add a `-> None` type hint?
</blockquote></details>

Done.
___
*[`silnlp/nmt/quality_estimation.py` line 40 at r3](https://reviewable.io/reviews/sillsdev/silnlp/793#-OXyneQdAPGZnRdWXf6B:-OYE4F4A7zcs4ZwEyjEO:b-896fix) ([raw file](https://github.com/sillsdev/silnlp/blob/4efd39b802831aeb00c07313899054c2c79bcd04/silnlp/nmt/quality_estimation.py#L40)):*

Can you refactor this so that we only do the regression once?  Linear regression isn't very expensive, but we may want to support other types of regression that require numerical or iterative methods someday.
</blockquote></details>

Done.
___
*[`silnlp/nmt/quality_estimation.py` line 205 at r3](https://reviewable.io/reviews/sillsdev/silnlp/793#-OXyfQhw47RywLEUz2vv:-OYDMFZI4YKZfXfiVm2U:b-896fix) ([raw file](https://github.com/sillsdev/silnlp/blob/4efd39b802831aeb00c07313899054c2c79bcd04/silnlp/nmt/quality_estimation.py#L205)):*

Very minor, but the abbreviation for 1 John is `1JN`.
</blockquote></details>

Done.
___
*[`silnlp/nmt/quality_estimation.py` line 208 at r3](https://reviewable.io/reviews/sillsdev/silnlp/793#-OXyiZA16mWL0gPW44zk:-OYDR6qiAOCAU1qTg8Am:b-9kym8b) ([raw file](https://github.com/sillsdev/silnlp/blob/4efd39b802831aeb00c07313899054c2c79bcd04/silnlp/nmt/quality_estimation.py#L208)):*

These two arguments also use different cases (snake vs. kebab)
</blockquote></details>

I think this is the intended casing for both arguments, since after parsing they should both be snake case. For positional arguments, the convention is to use snake case since the variable name is taken as is by the arg parser. For optional flags, the convention is to use kebab case so it's consistent with the -- in front, and arg parser automatically converts it to snake case.
___
*[`silnlp/nmt/quality_estimation.py` line 247 at r3](https://reviewable.io/reviews/sillsdev/silnlp/793#-OXylILhBKsS21FhwzOz:-OYDiOEs0ApaWMyfTcTM:bl1sq5i) ([raw file](https://github.com/sillsdev/silnlp/blob/4efd39b802831aeb00c07313899054c2c79bcd04/silnlp/nmt/quality_estimation.py#L247)):*

I believe this could result in unexpected behavior when we are either producing multiple drafts or applying postprocessing, since there could be multiple files that match this glob for each book.
</blockquote></details>

Would postprocessing files be included here? I didn't think any files with a postprocessing suffix also had a .confidences.tsv suffix.

For the multiple drafts case, I've added a --draft-index arg that can be used when --books is specified.
___
*[`silnlp/nmt/quality_estimation.py` line 249 at r3](https://reviewable.io/reviews/sillsdev/silnlp/793#-OXyo30WDaNRbnjtIdCn:-OYIyK_fEWLqYbl-_k3C:b-896fix) ([raw file](https://github.com/sillsdev/silnlp/blob/4efd39b802831aeb00c07313899054c2c79bcd04/silnlp/nmt/quality_estimation.py#L249)):*

I can imagine the diff predictions file being in a different experiment folder than the confidence files, since it will usually have different training data (holding out a random 100/250 verses)
</blockquote></details>

Done.

PR #801, opened 2025-09-03
Prevent model from being loaded multiple times if translate method is called multiple times

mshannon-sil:
Previously, if the translate method was called multiple times, such as when there is a list of books to translate, the model was loaded each time. The model is now saved as an instance variable as a form of caching, as are the translate parameters. Whenever a call to translate uses the same parameters (ckpt, src_lang, trg_lang) for the translation model, the model is reused. If the parameters are different, then a new inference model is created and replaces the old cached model.

### DIFF
--- a/silnlp/nmt/hugging_face_config.py
+++ b/silnlp/nmt/hugging_face_config.py
@@ -840,6 +840,8 @@ class HuggingFaceNMTModel(NMTModel):
         self._dictionary: Optional[Dict[VerseRef, Set[str]]] = None
         self._is_t5 = self._config.model_prefix in SUPPORTED_T5_MODELS
         self._num_devices = num_devices
+        self._cached_model: Optional[PreTrainedModel] = None
+        self._cached_translate_params: Optional[Tuple[Union[CheckpointType, str, int], str, str]] = None
 
     def train(self) -> None:
         training_args = self._create_training_arguments()
@@ -1218,7 +1220,11 @@ class HuggingFaceNMTModel(NMTModel):
         src_lang = self._config.data["lang_codes"].get(src_iso, src_iso)
         trg_lang = self._config.data["lang_codes"].get(trg_iso, trg_iso)
         tokenizer = self._config.get_tokenizer()
-        model = self._create_inference_model(ckpt, tokenizer, src_lang, trg_lang)
+        if self._cached_translate_params == (ckpt, src_lang, trg_lang) and self._cached_model is not None:
+            model = self._cached_model
+        else:
+            model = self._cached_model = self._create_inference_model(ckpt, tokenizer, src_lang, trg_lang)
+            self._cached_translate_params = (ckpt, src_lang, trg_lang)
         if model.config.max_length is not None and model.config.max_length < 512:
             model.config.max_length = 512
 


mshannon-sil:
*[`silnlp/nmt/hugging_face_config.py` line 1223 at r1](https://reviewable.io/reviews/sillsdev/silnlp/801#-OZF0EcA0LRswyhP9Gtn:-OZRdTfHDYwjXik9NbpC:b-tf476) ([raw file](https://github.com/sillsdev/silnlp/blob/39c74914c8f4f39cc2787d581e89b666a0323078/silnlp/nmt/hugging_face_config.py#L1223)):*

It would be good to clear the cached model, once we have finished translating all of the books. This isn't strictly necessary right now, but it might help us to avoid issues in the future. Here is how I would suggest doing it:

1. Add a `clear_cache` method to the config class.
2. Update the `Translator`/`NMTTranslator` class to be a context manager using `AbstractContextManager`. Call `clear_cache` in the `__exit__` method.
3. In `TranslationTask`, use `with` blocks whenever we construct a translator. This will ensure that the cache is cleared when we are done translating.
</blockquote></details>

This sounds good, and I've started working on it. Can you just confirm whether you'd prefer the `clear_cache` method in the `HuggingFaceConfig` class or the `HuggingFaceNMTModel` class from a design standpoint? To put it in the config class, I think I'd either need to move the cached model to the config class and then do some refactoring to get/set the cached model in the translate method which is not in the config class, or keep the parameters in `HuggingFaceNMTModel` but pass in a `HuggingFaceNMTModel` instance as a parameter to `clear_cache` so it has access. It seems simpler on the surface to put `clear_cache` in `HuggingFaceNMTModel`, but if it's preferable for the config class to do everything related to resource management or for the `clear_cache` method to be more broadly applicable I can definitely move it there.

mshannon-sil (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/801)* status: 0 of 2 files reviewed, 2 unresolved discussions (waiting on @benjaminking, @ddaspit, and @mshannon-sil)

___
*[`silnlp/nmt/hugging_face_config.py` line 844 at r1](https://reviewable.io/reviews/sillsdev/silnlp/801#-OZFyNhqEfcGXcvT-rY4:-O_Uq7qTC2yzj-GzHgv3:bs7wumx) ([raw file](https://github.com/sillsdev/silnlp/blob/39c74914c8f4f39cc2787d581e89b666a0323078/silnlp/nmt/hugging_face_config.py#L844)):*

When we have complicated type signatures like this, I like to try to factor out a class to represent the type.  The class can also have logic that validates the type.
</blockquote></details>

Done. 
___
*[`silnlp/nmt/hugging_face_config.py` line 1223 at r1](https://reviewable.io/reviews/sillsdev/silnlp/801#-OZF0EcA0LRswyhP9Gtn:-O_UbCWJCsvwsmh4Y2Ze:b-896fix) ([raw file](https://github.com/sillsdev/silnlp/blob/39c74914c8f4f39cc2787d581e89b666a0323078/silnlp/nmt/hugging_face_config.py#L1223)):*

You are correct. The `clear_cache` method should added to the `HuggingFaceNmtModel` class.
</blockquote></details>

Done.

benjaminking (approved):
:lgtm:

> @benjaminking reviewed 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/801#-:-O_Xh9zIDPXyR37nS2b0:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @ddaspit)

ddaspit (changes requested):
> @ddaspit reviewed 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/801)* status: all files reviewed, 1 unresolved discussion (waiting on @mshannon-sil)

___
*[`silnlp/nmt/translate.py` line 41 at r2](https://reviewable.io/reviews/sillsdev/silnlp/801#-O_YPjD5C1WkvOhEXot8:-O_YPjD5C1WkvOhEXot9:bu76cvc) ([raw file](https://github.com/sillsdev/silnlp/blob/3a709d513747da0d1cc4d545b5d4e7b8e6a3df16/silnlp/nmt/translate.py#L41)):*
> ```Python
>     
>     def __exit__(self, exc_type, exc_value, traceback) -> None:
>         if hasattr(self._model, 'clear_cache'):
> ```

You should add `clear_cache` to the `NMTModel` interface. That way you don't need to perform this check.

mshannon-sil (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/801)* status: 1 of 3 files reviewed, 1 unresolved discussion (waiting on @benjaminking and @ddaspit)

___
*[`silnlp/nmt/translate.py` line 41 at r2](https://reviewable.io/reviews/sillsdev/silnlp/801#-O_YPjD5C1WkvOhEXot8:-O_mDQYG5zfuN-8wS60O:b-896fix) ([raw file](https://github.com/sillsdev/silnlp/blob/3a709d513747da0d1cc4d545b5d4e7b8e6a3df16/silnlp/nmt/translate.py#L41)):*

You should add `clear_cache` to the `NMTModel` interface. That way you don't need to perform this check.
</blockquote></details>

Done.

PR #806, opened 2025-09-05
Quotation denormalization with machine

benjaminking:
This PR implements quotation mark denormalization using Machine.py.  This required a fair bit of refactoring because of the requirement that the quote conventions of the source and target projects be able to be automatically detected, if the user does not specify them.  This means that information about the corpora used for training and translation needs to be passed to `translate_usfm()`.  Incidentally, some of the command line script parameters had to be modified to support this, since a USFM file with no context doesn't have enough information to have the source/target quote convention detected.  I've also made some changes to have SILNLP's postprocessing better match Serval's postprocessing.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/806)

### DIFF
--- a/silnlp/common/postprocesser.py
+++ b/silnlp/common/postprocesser.py
@@ -187,8 +187,8 @@ class DenormalizeQuotationMarksPostprocessor:
             raise NoDetectedQuoteConventionException(project_name)
         LOGGER.info(
             f'Detected quote convention for project "{project_name}" is '
-            + '"{quote_convention_analysis.best_quote_convention.name}" with score '
-            + "{quote_convention_analysis.best_quote_convention_score:.2f}."
+            + f'"{quote_convention_analysis.best_quote_convention.name}" with score '
+            + f"{quote_convention_analysis.best_quote_convention_score:.2f}."
         )
         self._project_convention_cache[project_name] = quote_convention_analysis.best_quote_convention
 
@@ -222,17 +222,7 @@ class DenormalizeQuotationMarksPostprocessor:
 
         if len(processed_chapters) == 0:
             return self._NO_CHAPTERS_REMARK_SENTENCE
-        return (
-            self._REMARK_SENTENCE
-            + ", ".join(
-                [
-                    str(chapter_num)
-                    for chapter_num, strategy in enumerate(best_chapter_strategies, 1)
-                    if strategy != QuotationMarkUpdateStrategy.SKIP
-                ]
-            )
-            + "."
-        )
+        return self._REMARK_SENTENCE + ", ".join(processed_chapters) + "."
 
     def postprocess_usfm(
         self,
--- a/silnlp/nmt/config.py
+++ b/silnlp/nmt/config.py
@@ -15,19 +15,6 @@ from machine.scripture import ORIGINAL_VERSIFICATION, VerseRef, get_books
 from machine.tokenization import LatinWordTokenizer
 from tqdm import tqdm
 
-from silnlp.common.translator import TranslationGroup
-from silnlp.nmt.corpora import (
-    BASIC_DATA_PROJECT,
-    CorpusPair,
-    DataFile,
-    DataFileMapping,
-    IsoPairInfo,
-    get_data_file_pairs,
-    get_parallel_corpus_size,
-    get_terms_glosses_file_paths,
-    parse_corpus_pairs,
-)
-
 from ..alignment.config import get_aligner_name
 from ..alignment.utils import add_alignment_scores
 from ..common.corpus import (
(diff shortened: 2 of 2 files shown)


Enkidu93 (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/806)* status: all files reviewed, 5 unresolved discussions (waiting on @benjaminking and @mshannon-sil)

___
*[`silnlp/nmt/postprocess.py` line 5 at r1](https://reviewable.io/reviews/sillsdev/silnlp/806#-OZdNW4Y52JNbT5wH3S4:-OZdNW4Y52JNbT5wH3S5:b-g2yyj2) ([raw file](https://github.com/sillsdev/silnlp/blob/0de38e8182b03ae74b040d29263d5a1c2da6bb74/silnlp/nmt/postprocess.py#L5)):*
> ```Python
> import re
> from pathlib import Path
> from typing import Dict, List, Optional
> ```

I think `Dict` is unused. 
___
*[`silnlp/nmt/postprocess.py` line 147 at r1](https://reviewable.io/reviews/sillsdev/silnlp/806#-OZdNt8l0UYOxVyOkMKU:-OZdNt8n4_eRlZf53a35:b-p4bsp0) ([raw file](https://github.com/sillsdev/silnlp/blob/0de38e8182b03ae74b040d29263d5a1c2da6bb74/silnlp/nmt/postprocess.py#L147)):*
> ```Python
>         if src_sentence.ref.to_relaxed() != draft_sentence.ref.to_relaxed():
>             LOGGER.warning(
>                 f"Can't process {draft_metadata.source_path} and {draft_path}: Mismatched ref, {src_ref} != {draft_ref}. Files must have the exact same USFM structure"
> ```

I think `draft_path` is supposed to be `draft_metdata.draft_path` - a couple similar issues in these log messages.

mshannon-sil (changes requested):
> @mshannon-sil reviewed 2 of 10 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/806)* status: 8 of 10 files reviewed, 9 unresolved discussions (waiting on @benjaminking and @Enkidu93)

___
*[`silnlp/common/postprocesser.py` line 21 at r2](https://reviewable.io/reviews/sillsdev/silnlp/806#-OZp2Zn4BBgtYFu4X-KY:-OZp2Zn4BBgtYFu4X-KZ:baj2i0y) ([raw file](https://github.com/sillsdev/silnlp/blob/bc3e5c16a1076cccd570593f941f4a7100e4badb/silnlp/common/postprocesser.py#L21)):*
> ```Python
>     parse_usfm,
> )
> from machine.punctuation_analysis import (
> ```

Reminder that we need to update sil-machine in pyproject.toml to the latest version and update the lock file with `poetry lock --no-update` for this to work (I know that it hadn't been released when this was originally written).
___
*[`silnlp/common/postprocesser.py` line 105 at r2](https://reviewable.io/reviews/sillsdev/silnlp/806#-OZpFPFw0MCkTlvTX2vU:-OZpFPFx36cd3qKLAx1U:b-oi007t) ([raw file](https://github.com/sillsdev/silnlp/blob/bc3e5c16a1076cccd570593f941f4a7100e4badb/silnlp/common/postprocesser.py#L105)):*
> ```Python
>         usfm: str,
>         rows: List[UpdateUsfmRow],
>         remarks: List[str] = [],
> ```

I learned this recently, but it's not recommended to have mutable default arguments in Python. Default values are only evaluated once, so if the default is used and then modified in a function call, if the function is called another time, the default remains modified. It's better to set the default to None, and then initialize the argument at the beginning of the method if the argument is None. It would be good to go through your code and modify all the places where mutable default arguments ([], {}, etc.) are being used.
___
*[`silnlp/common/postprocesser.py` line 160 at r2](https://reviewable.io/reviews/sillsdev/silnlp/806#-OZpGkDhDX2n9wkc1rCA:-OZpGkDi5VC19g4M7r6s:bpygs2o) ([raw file](https://github.com/sillsdev/silnlp/blob/bc3e5c16a1076cccd570593f941f4a7100e4badb/silnlp/common/postprocesser.py#L160)):*
> ```Python
>                     "The source project name must be explicitly provided or be present in translate_config.yml, since an explicit source quote convention name was not provided."
>                 )
>             if selected_training_books is None:
> ```

If {} is passed in for selected_training_books, this will not catch it.
___
*[`silnlp/common/postprocesser.py` line 233 at r2](https://reviewable.io/reviews/sillsdev/silnlp/806#-OZpHVrq5E_TBrE36sQu:-OZpHVrq5E_TBrE36sQv:b-xfkipn) ([raw file](https://github.com/sillsdev/silnlp/blob/bc3e5c16a1076cccd570593f941f4a7100e4badb/silnlp/common/postprocesser.py#L233)):*
> ```Python
>                 ]
>             )
>             + "."
> ```

I think as written if every chapter is skipped it results in `Quotation marks in the following chapters have been automatically denormalized after translation: .` Doesn't matter too much so this comment isn't blocking but thought I would note it.
___
*[`silnlp/common/postprocesser.py` line 287 at r2](https://reviewable.io/reviews/sillsdev/silnlp/806#-OZpBOf27G1xuzjN4Q8G:-OZpBOf27G1xuzjN4Q8H:b8xieem) ([raw file](https://github.com/sillsdev/silnlp/blob/bc3e5c16a1076cccd570593f941f4a7100e4badb/silnlp/common/postprocesser.py#L287)):*
> ```Python
>             if option in POSTPROCESS_SUFFIX_CHARS and self._config[option] != default:
>                 if isinstance(default, str):
>                     suffix += POSTPROCESS_SUFFIX_CHARS[option][self._config[option]]
> ```

Should the if else statements here be the other way around? I would think that if the default is a string, then wouldn't `POSTPROCESS_SUFFIX_CHARS[option]` return a string not a dict?

PR #811, opened 2025-09-10
Fix issue with translate from saved checkpoint

TaperChipmunk32:
This PR fixes an issue with translate using the base model even when a checkpoint is present.

### DIFF
--- a/silnlp/nmt/clearml_connection.py
+++ b/silnlp/nmt/clearml_connection.py
@@ -123,9 +123,9 @@ class SILClearML:
         if self.task is None:
             with (exp_dir / "config.yml").open("r", encoding="utf-8") as file:
                 config = yaml.safe_load(file)
-                config["use_default_model_dir"] = self.use_default_model_dir
             if config is None or len(config.keys()) == 0:
                 raise RuntimeError("Config file has no contents.")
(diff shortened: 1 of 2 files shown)


benjaminking (changes requested):
> @benjaminking reviewed 2 of 2 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/811)* status: all files reviewed, 1 unresolved discussion (waiting on @mshannon-sil)

___
*[`silnlp/nmt/translate.py` line 404 at r1](https://reviewable.io/reviews/sillsdev/silnlp/811#-OZpXQV8D5Bdd3rAfuQi:-OZpXQV8D5Bdd3rAfuQj:b-vzo3eu) ([raw file](https://github.com/sillsdev/silnlp/blob/500b6684ecdf4f807ebf20633d94717c61df6e8d/silnlp/nmt/translate.py#L404)):*
> ```Python
>         name=args.experiment,
>         checkpoint=args.checkpoint,
>         save_checkpoints=True,
> ```

Could you rename this parameter (just for `TranslationTask`)?  I feel like it's a little misleading here, since you're not saving checkpoints during translation and in fact, you may have it be true even when no checkpoints are saved, for example if you want to inference from the base model.

PR #848, opened 2025-10-27
Improvements to onboard_project

TaperChipmunk32:
- Can now onboard multiple projects
- `--copy-from` now will use the user's Downloads folder if included with no value 
- clean_project changed from `silnlp.scripts.clean_project` to `silnlp.common.clean_projects` so multiple projects can be cleaned at once and more unnecessary files are removed 
- Added a wildebeest section to the config, and default args for wildebeest analysis

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/848)

### DIFF
--- a/silnlp/common/onboard_project.py
+++ b/silnlp/common/onboard_project.py
@@ -167,12 +167,6 @@ def main() -> None:
     parser.add_argument(
         "--wildebeest", default=False, action="store_true", help="Run Wildebeest analysis on the extracted corpora."
     )
-    parser.add_argument(
-        "--zip-password",
-        default=None,
-        type=str,
-        help="Password for a Paratext project zip file, if it is encrypted.",
-    )
 
     args = parser.parse_args()
     if not args.projects:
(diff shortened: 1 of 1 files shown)


TaperChipmunk32 (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/848)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @benjaminking and @mmartin9684-sil)

___
*[`silnlp/common/onboard_project.py` line 170 at r1](https://reviewable.io/reviews/sillsdev/silnlp/848#-OcbNCW_2P96Lxpr0yXR:-Ocf4gzYALfqA7A_aRtX:b-ik9tfq) ([raw file](https://github.com/sillsdev/silnlp/blob/a61fc00c14decadaae656e1fb68ce78f054d796d/silnlp/common/onboard_project.py#L170)):*

How does this work when the user passes multiple projects?  Can they pass multiple zip files with different passwords?
</blockquote></details>

No, it would not work. I have changed this to instead use a `zip_password` section in the config. The config would look something like this:



_Code snippet:_
```Python
zip_password:
    project_name_1: password_1
    project_name_2: password_2
```
___
*[`silnlp/common/onboard_project.py` line 200 at r1](https://reviewable.io/reviews/sillsdev/silnlp/848#-OcbNjE44s6nVJkUE71n:-Ocf4qAp2s3WR03KfvCd:bhc0aek) ([raw file](https://github.com/sillsdev/silnlp/blob/a61fc00c14decadaae656e1fb68ce78f054d796d/silnlp/common/onboard_project.py#L200)):*

I'm guessing that this change was driven by a request from the EITL team, but I liked prompting the user for the passwords, to help keep passwords out of the command history.  But if EITL requested it to work this way, we should go with that.
</blockquote></details>

Yes, because the script may take some time to run, it would be inconvenient to have to watch for any password prompts.

PR #883, opened 2025-12-02
Verse segmentation fix

laura-burdick-sil:
Previously, the verse segmentation algorithm was crashing when there was a passage that was only a single verse. This fixes that behavior.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/883)


laura-burdick-sil:
I've also added some code to output the resulting verses in vref format (so that they can easily be fed directly into NLLB). @Enkidu93, would you also be able to review this code? Thank you!

PR #885, opened 2025-12-03
Changes update_src and update_trg default to true

TaperChipmunk32:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/885)


mshannon-sil:
@ddaspit that's true, we should communicate that to the team. It's probably worth it in this case, as this will bring SILNLP in line with the default behavior of Serval.

PR #926, opened 2026-01-30
Use machine for key terms extraction

Enkidu93:
This PR replaces silnlp's key term ad hoc extraction code with calls to the machine library.

Not sure who all should be pinged for review!

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/926)


Enkidu93 (commented):
> @Enkidu93 made 3 comments.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/926)* status: all files reviewed, 3 unresolved discussions (waiting on @benjaminking and @ddaspit).

___
*[`silnlp/common/paratext.py` line 232 at r1](https://reviewable.io/reviews/sillsdev/silnlp/926#-OkFFQh34ZrXC73qpmcf:-OkZczuI5jOeQyCGvNTb:buawqho) ([raw file](https://github.com/sillsdev/silnlp/blob/600072d55980b68ae6c54a9cf1c3bf9d2efa848b/silnlp/common/paratext.py#L232)):*

Could you use a type alias or maybe a wrapper class for the return value here?  It's not obvious from reading the signature what is going to be returned.  If it just said `-> TermsCount:` or something like that, it would make it easier to understand.
</blockquote></details>

That's a good idea. Are you ok with a docstring? If so, done.
___
*[`silnlp/common/paratext.py` line 241 at r1](https://reviewable.io/reviews/sillsdev/silnlp/926#-OkFK2IH-IE5n6SP7qwq:-OkZeB2z3vQts7yXMBr7:b-dixn4t) ([raw file](https://github.com/sillsdev/silnlp/blob/600072d55980b68ae6c54a9cf1c3bf9d2efa848b/silnlp/common/paratext.py#L241)):*

It doesn't look like you use this path after checking that it exists.  Could you add a comment explaining why you need to make this check?
</blockquote></details>

Done. Internally machine.py accesses this path. I added a comment: Mainly, we just want to avoid extra processing if we don't have to. If you'd rather, I can remove it.
___
*[`silnlp/nmt/config.py` line 144 at r1](https://reviewable.io/reviews/sillsdev/silnlp/926#-OkVAIAd7a2zq3mUKgiG:-OkZf5ML74q9KPT8hIKT:b-896fix) ([raw file](https://github.com/sillsdev/silnlp/blob/600072d55980b68ae6c54a9cf1c3bf9d2efa848b/silnlp/nmt/config.py#L144)):*

*turn
</blockquote></details>

Done.

PR #936, opened 2026-02-11
Further refine confidence files and quality estimation

mshannon-sil:
This PR does the following:

- Improves the encapsulation for the ConfidenceFile class, especially by adding iterators for all types of confidence files and moving some "write confidence scores to file" methods from the TranslatedDraft class to the ConfidenceFile class
- Creates the UsfmConfidenceFile, TxtConfidenceFile, and TestConfidenceFile subclasses to provide the appropriate properties and methods for each type of confidence file.
- FIxes an issue where book confidence was being appended to confidences.books.tsv even when an entry for that book was already present
- Moves the quality-estimation and verse-test-scores-file arguments from experiment.py to translate_config.yml
- Updates quality_estimation.py to use the new ConfidenceFile structure and subclasses
- Added an auto-detect mode to quality_estimation.py to run quality estimation on all confidence files in a directory
- Added the ability to do quality estimation on a TxtConfidenceFile
- Updates the green/yellow threshold to an equivalent of 50 chrF3
- Created a constant for the test trg-predictions prefix, with the goal of using the constant in the ConfidenceFile class, but that resulted in circular imports. I left the constant in since it might still be helpful for test.py, but I can remove it if that's preferred

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/936)

### DIFF
--- a/silnlp/nmt/quality_estimation.py
+++ b/silnlp/nmt/quality_estimation.py
@@ -436,22 +436,26 @@ def main() -> None:
     parser.add_argument(
         "verse_test_scores_file",
         type=str,
-        help="The tsv file relative to MT/experiments containing the verse-level test scores to determine "
-        + "line of best fit, e.g., `project_folder/exp_folder/test.trg-predictions.detok.txt.5000.scores.tsv`. ",
+        help="Path relative to MT/experiments to a verse-level test score file, which is used to find line of best fit "
+        + "for confidence and chrF3, e.g., project_folder/exp_folder/test.trg-predictions.detok.txt.5000.scores.tsv. "
+        + "If a directory is provided instead, the first *.scores.tsv match is used.",
     )
     parser.add_argument(
         "confidence_files",
         nargs="*",
         type=str,
-        help="Relative paths for the confidence files to process (relative to MT/experiments or --confidence-dir "
-        + "if specified) e.g. 'project_folder/exp_folder/infer/5000/source/631JN.SFM.confidences.tsv' or "
-        + "'631JN.SFM.confidences.tsv --confidence-dir project_folder/exp_folder/infer/5000/source'.",
+        help="Zero or more confidence file paths (.confidences.tsv suffix, e.g., "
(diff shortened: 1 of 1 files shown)


benjaminking (approved):
:lgtm:

> @benjaminking reviewed 1 file and all commit messages, made 2 comments, and resolved 3 discussions.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/936#-:-Oll9JwR0DGz4kOjDUjR:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on ddaspit, mshannon-sil, and TaperChipmunk32).

___
*[`silnlp/nmt/quality_estimation.py` line 64 at r2](https://reviewable.io/reviews/sillsdev/silnlp/936#-OlNdIT79qgAIquouOZQ:-Oll6Ukg5KrQYWvPY8RU:b-i6l7l2) ([raw file](https://github.com/sillsdev/silnlp/blob/cb8b57d12bd1e15d8c353d1a36aa3eb9a46c0fb6/silnlp/nmt/quality_estimation.py#L64)):*

In the project_chrf3 method, it iterates through a list of ConfidenceFile items. If it's a UsfmConfidenceFile, then each instance has a chapter and book confidence file associated with it. The issue is that they're structured differently; there is an individual chapter confidence file that corresponds to each confidence file, but the book confidence file is a shared file for every usfm confidence file in that directory (to make comparison easy and reduce the clutter). So as it goes through each confidence file to do quality estimation, the add_score method for BookScores has a way to check if it's already seen that book confidence file and if so to skip it.

I could just only add one score at a time rather than all the scores in the file, to avoid needing this check, but I thought it would be inefficient if we'd already be iterating through the file anyway. Also, I think most of the time there should only be one book confidence file since there can only be one such file in a directory, which if that were the only scenario would allow me to take it out of the loop and again maybe not need this check. But then it would break in the edge case of doing quality estimation on confidence files in multiple directories.
</blockquote></details>

It should be fine the way it's written.  I was unsure why we'd have duplicates, but it makes sense now.

PR #941, opened 2026-02-16
Copy project folder when onboarding rather than renaming

TaperChipmunk32:
The onboard_project script will now copy the PT project to a new folder when renaming the path

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/941)

### DIFF
--- a/silnlp/common/onboard_project.py
+++ b/silnlp/common/onboard_project.py
@@ -2,6 +2,7 @@ import argparse
 import getpass
 import logging
 import re
+import shutil
 import sys
 import xml.etree.ElementTree as ET
 import zipfile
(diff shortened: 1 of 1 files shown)


benjaminking (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/941)* status: all files reviewed, 1 unresolved discussion (waiting on TaperChipmunk32).

___
*[`silnlp/common/onboard_project.py` line 309 at r1](https://reviewable.io/reviews/sillsdev/silnlp/941#-OllB0fM6b4_RkJGuGvS:-OllB0fN5D2sA3AVkkFM:bv9ssw0) ([raw file](https://github.com/sillsdev/silnlp/blob/7b4ee404b574cb8f6011fe0edefda2ecaccea86e/silnlp/common/onboard_project.py#L309)):*
> ```Python
>         if not new_local_project_path.exists():
>             new_local_project_path.mkdir(parents=True, exist_ok=True)
>         _copy_directory_to_paratext_project(local_project_path, new_local_project_path, overwrite=True)
> ```

If I'm remembering correctly, this was because folders were getting renamed in the user's local Paratext projects directory.  If we're creating a new folder in their projects directory, do we need to delete that folder when we're done?

Issue #929, opened 2026-02-03
Project versification needs to be read from the parent project for non-Standard Paratext projects


mmartin9684-sil:
For non-Standard Paratext projects, the project versification is not typically specified in the project's _Settings.xml_ file.  Instead, the versification needs to be read from the parent project's _Settings.xml_ file.  Since the versification is not found in the project's _Settings.xml_ file for non-Standard projects, silnlp current defaults to assigning English (4) versification to the project.  If the parent project uses a different versification (e.g., Original), the extract_corpora script may create an incorrect verse extract file; the behavior of other tools may also be negatively affected by this.

Paratext project types that do not typically have their own project versification include: Auxilliary; BackTranslation; ConsultantNotes; Daughter; Front Translation/FrontTranslation; StudyBible; Transliteration; TransliterationWithEncoder.

- There may be other project types as well.
- Some projects do not have a project type specified in their _Settings.xml_ file.

Our current method of storing multiple datestamp'ed versions of a parent project in the Paratext > projects folder makes it more challenging to match the parent project name from the _Settings.xml_ file to the parent project folder.

Another difficulty is that we do not typically download and onboard the parent project when working with these non-Standard projects.  This workflow will need to be updated so that the parent project's versification setting is available.

A partial refresh of the projects and project extracts for the affected non-Standard projects may be needed.

Issue #875, opened 2025-11-21
Utilize machine.py's `UsfmVersificationErrorDetector`


Enkidu93:
This is already being used in Serval to generate warnings for developers and Serval admins. 

We could integrate this into silnlp as a standalone script or maybe as part of `extract_corpora`.

(See https://github.com/sillsdev/machine.py/pull/245 for code updates).

PR #946, opened 2026-02-19
Properly support daughter translations

Enkidu93:
Also, add versification error detection to extraction

Requires https://github.com/sillsdev/machine.py/pull/267

Fixes https://github.com/sillsdev/silnlp/issues/929, https://github.com/sillsdev/silnlp/issues/875

@TaperChipmunk32, let me know if there's a better way to incorporate these updates into `onboard_project`!

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/946)

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -5178,22 +5178,24 @@ type = ["importlib-metadata (>=7.0.2)", "jaraco.develop (>=7.21)", "mypy (==1.12
 
 [[package]]
 name = "sil-machine"
-version = "1.8.6"
+version = "1.8.7"
 description = "A natural language processing library that is focused on providing tools for resource-poor languages."
 optional = false
-python-versions = ">=3.10,<3.14"
-files = []
-develop = true
+python-versions = "<3.14,>=3.10"
+files = [
+    {file = "sil_machine-1.8.7-py3-none-any.whl", hash = "sha256:c6fdd64b0a071bdeb9ca6d999d3685b15d0ff5d9104758eb90e54ab37dc1d885"},
+    {file = "sil_machine-1.8.7.tar.gz", hash = "sha256:5a6e7830ad609ec3b0c1b9c6bb1eedd22b63c3eb28493d2c6c18790439f2a654"},
+]
 
 [package.dependencies]
-boto3 = "^1.19.41"
-botocore = "^1.35.41"
-charset-normalizer = "^2.1.1"
-networkx = "^3"
-numpy = "^2.2.6"
+boto3 = ">=1.19.41,<2.0.0"
+botocore = ">=1.35.41,<2.0.0"
+charset-normalizer = ">=2.1.1,<3.0.0"
+networkx = ">=3,<4"
+numpy = ">=2.2.6,<3.0.0"
 regex = ">=2021.7.6"
-sil-thot = "^3.4.9"
-sortedcontainers = "^2.4.0"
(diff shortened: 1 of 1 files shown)


ddaspit (commented):
> @ddaspit reviewed 2 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/946)* status: all files reviewed, 2 unresolved discussions (waiting on TaperChipmunk32).

___
*[`silnlp/common/paratext.py` line 45 at r1](https://reviewable.io/reviews/sillsdev/silnlp/946#-Olrxq-_8SyzL6-zt585:-OlwLobk6jU2ox3llZPO:b-ht3vaa) ([raw file](https://github.com/sillsdev/silnlp/blob/41383ac791be0ffd89557469c862cf4c84f4305b/silnlp/common/paratext.py#L45)):*

Since the loop doesn't take very long and speeding it up is not as straightforward as I thought it would be, I would go with your solution and not complicate it.
</blockquote></details>

A few seconds seems like a long time to me. I would like to see this optimized if possible.

Enkidu93 (commented):
> @Enkidu93 made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/946)* status: 5 of 6 files reviewed, 2 unresolved discussions (waiting on benjaminking and TaperChipmunk32).

___
*[`silnlp/common/paratext.py` line 45 at r1](https://reviewable.io/reviews/sillsdev/silnlp/946#-Olrxq-_8SyzL6-zt585:-Olwn3mwAshuVFGtfCFN:bz7jjp) ([raw file](https://github.com/sillsdev/silnlp/blob/41383ac791be0ffd89557469c862cf4c84f4305b/silnlp/common/paratext.py#L45)):*

A few seconds seems like a long time to me. I would like to see this optimized if possible.
</blockquote></details>

OK! Sorting by datestamp slowed things down, so unless we want to ensure that it always finds the most recently updated parent (if there are multiple copies of the parent project), I don't think we should do that. If you want to explicitly point to a certain copy of the parent project, I added a parent project command-line argument.

Looking first by parent project name of course sped things up if the directory does contain the parent name. It still takes 1.5-4 seconds with my test projects. For reference, the rest of the extract script (which can vary a lot depending on the size of the project) takes something on the order of 10 seconds.

ddaspit (approved):
:lgtm:

> @ddaspit reviewed 5 files and all commit messages, and made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/946#-:-OmFNqqy4uPYLbbJNrvj:bnfp4nl)* status: all files reviewed, 2 unresolved discussions (waiting on ddaspit and TaperChipmunk32).

___
*[`silnlp/common/paratext.py` line 45 at r1](https://reviewable.io/reviews/sillsdev/silnlp/946#-Olrxq-_8SyzL6-zt585:-OmFMwt5E0if5JCFjdhN:b-do1061) ([raw file](https://github.com/sillsdev/silnlp/blob/41383ac791be0ffd89557469c862cf4c84f4305b/silnlp/common/paratext.py#L45)):*

None of the projects I tested had a parent whose project folder was exactly its name. Often they have a date suffixed on them, sometimes the capitalization is different, or otherwise, they can just not exactly match. But I do think it's a good idea to check before resorting to scanning the whole directory. Done.
</blockquote></details>

That's unfortunate. This looks like the best that we can do.

Issue #947, opened 2026-02-22
experiment.py: --save-confidences does not support multilingual drafting configurations


mmartin9684-sil:
If the --save-confidences argument is used on the experiment script with a multilingual training configuration, the script will fail.  For instance, this corpus_pair configuration:
```
  - corpus_books: GEN;EXO;NT
    mapping: many_to_many
    src:
    - tpi-TPB12
    - en-NIrV
    trg: ena-AplDupl_2025_11_29
    type: train,test
```
will fail during the test stage because it does not handle multilingual predictions files for the test set (i.e., test._src_._trg_.trg-predictions.txt._checkpoint_).

A sample stack trace is provided below.
```
100% 250/250 [01:13<00:00,  3.41ex/s]
Traceback (most recent call last):
  File "/usr/lib/python3.10/runpy.py", line 196, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/usr/lib/python3.10/runpy.py", line 86, in _run_code
    exec(code, run_globals)
  File "/root/.clearml/venvs-builds/3.10/task_repository/silnlp.git/silnlp/nmt/experiment.py", line 295, in <module>
    main()
  File "/root/.clearml/venvs-builds/3.10/task_repository/silnlp.git/silnlp/nmt/experiment.py", line 291, in main
    exp.run()
  File "/root/.clearml/venvs-builds/3.10/task_repository/silnlp.git/silnlp/nmt/experiment.py", line 58, in run
    self.test()
  File "/root/.clearml/venvs-builds/3.10/task_repository/silnlp.git/silnlp/nmt/experiment.py", line 82, in test
    test(
  File "/root/.clearml/venvs-builds/3.10/task_repository/silnlp.git/silnlp/nmt/test.py", line 820, in test
    results[step] = test_checkpoint(
  File "/root/.clearml/venvs-builds/3.10/task_repository/silnlp.git/silnlp/nmt/test.py", line 595, in test_checkpoint
    model.translate_test_files(
  File "/root/.clearml/venvs-builds/3.10/task_repository/silnlp.git/silnlp/nmt/hugging_face_config.py", line 1218, in translate_test_files
    generate_confidence_files(
  File "/root/.clearml/venvs-builds/3.10/task_repository/silnlp.git/silnlp/common/translator.py", line 402, in generate_confidence_files
    confidence_file = ConfidenceFile.from_draft_file_path(trg_draft_file_path)
[...]

Issue #940, opened 2026-02-13
High RAM usage during test step


mshannon-sil:
On multiple clearml experiments over the past week, we've noticed that if experiments are run using --preprocess and --test without the --train step, RAM utilization spikes to over 200GB, causing experiments to crash. We need to investigate what in particular during the test step is causing this drastic increase in RAM usage.

PR #960, opened 2026-03-06
fix spbleu ram issue; loosen check for TestConfidenceFile

mshannon-sil:
This PR fixes two bugs.

The first is #940, where the spbleu metric was reinitializing the Flores200 tokenizer on every iteration in the loop and consequently eating up all the RAM to the point of failure. I have moved intialization outside the loop to prevent that. The memory issue has gone away, and it has also had the side effect of drastically increasing the speed of the test step when spbleu is included. 

The second is #947, where multilingual drafting was not compatible with saving confidence scores. This was only due to the fact that the code to check for a valid filename for a TestConfidenceFile was too restrictive and did not account for filenames where the lang id is inserted toward the beginning. I have loosened the check just to look for the presence of "trg-predictions" in the trg draft file name and confirmed multilingual drafting experiments now complete successfully.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/960)

### DIFF
--- a/silnlp/common/translator.py
+++ b/silnlp/common/translator.py
@@ -49,7 +49,7 @@ class ConfidenceFile(ABC, Generic[TVerseKey]):
 
     @classmethod
     def _get_confidence_file_type(cls, trg_draft_file_path: Path) -> type["ConfidenceFile"]:
-        if trg_draft_file_path.name.startswith("test.trg-predictions"):
+        if "trg-predictions" in trg_draft_file_path.name:
             return TestConfidenceFile
         ext = trg_draft_file_path.suffix.lower()
         if ext in {".usfm", ".sfm"}:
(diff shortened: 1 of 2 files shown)


TaperChipmunk32:
*[`silnlp/nmt/test.py` line 267 at r1](https://reviewable.io/reviews/sillsdev/silnlp/960#-On4EZ1jAEqPV_XOoIp3:-On4EZ1jAEqPV_XOoIp4:b-mquslj) ([raw file](https://github.com/sillsdev/silnlp/blob/bb2b13a5cd0af3af12c646e551dbd3bd40fa8e08/silnlp/nmt/test.py#L267)):*
> ```Python
>     scorers = scorers.intersection(SUPPORTED_SENTENCE_SCORERS)
>     other_scores = {k: v for k, v in other_scores.items() if k.lower() in scorers}
>     other_verse_scores: Dict[str, float] = {}
> ```

Why was this removed? It seems to still be needed later in this function.

mshannon-sil:
*[`silnlp/nmt/test.py` line 300 at r1](https://reviewable.io/reviews/sillsdev/silnlp/960#-On4G86mESZ3EDJOgSQe:-On4GTB471Zt4CFtXh0R:bzfog5g) ([raw file](https://github.com/sillsdev/silnlp/blob/1579d2f91e3d94b89ab12cbed68b43f9db861b21/silnlp/nmt/test.py#L300)):*

Why was this changed?
</blockquote></details>

When I was making the fixes, I noticed that a couple scorers were still using corpus level metrics, even though these are sentence level scores. So I adjusted those scorers to use their sentence-level rather than corpus-level scorers.

PR #987, opened 2026-04-02
Multilingual inference (separate pipeline for each language pair)

benjaminking:
This is a revised version of the earlier multilingual inference PR, but this version is more limited in its scope, assuming that every test file will represent a single language pair.  The major changes are

* A new `SilTranslationPipeline` is created for each test file
* If the language codes are part of the test file name (e.g. `test.en.fr.src.txt`), the language codes will be extracted from the file name
* Otherwise, they will come from the model's config
* The method `_configure_model` is now called for saved models (previously was only called for non-finetuned models and MADLAD)
* The model's generation config is explicitly passed to the `generate` method (this keeps Huggingface from overwriting the generation config during the pipeline for obscure reasons)

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/987)

### DIFF
--- a/silnlp/nmt/hugging_face_config.py
+++ b/silnlp/nmt/hugging_face_config.py
@@ -1182,19 +1182,14 @@ class HuggingFaceNMTModel(NMTModel):
     ) -> None:
         tokenizer = self._config.get_tokenizer()
         model = self._create_inference_model(ckpt, tokenizer, self._config.test_src_lang, self._config.test_trg_lang)
-        pipeline = PretokenizedTranslationPipeline(
-            model=model,
-            tokenizer=tokenizer,
-            src_lang=self._config.test_src_lang,
-            tgt_lang=self._config.test_trg_lang,
-            device=0,
-        )
-        pipeline.model = torch.compile(pipeline.model)
+
         for input_path, translation_path, vref_path in zip(
             input_paths,
             translation_paths,
             cast(Iterable[Optional[Path]], repeat(None) if vref_paths is None else vref_paths),
         ):
+            pipeline = self._create_pipeline_for_test_file(input_path, model, tokenizer)
+
             length = count_lines(input_path)
             with ExitStack() as stack:
                 src_file = stack.enter_context(input_path.open("r", encoding="utf-8-sig"))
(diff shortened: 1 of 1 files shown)


mshannon-sil:
*[`silnlp/nmt/hugging_face_config.py` line 1243 at r1](https://reviewable.io/reviews/sillsdev/silnlp/987#-Or-qnfdBf0rEyiXfoD6:-Or-qnfdBf0rEyiXfoD7:bgdtnrf) ([raw file](https://github.com/sillsdev/silnlp/blob/5356d22e84f45d6d24b1a4a311a522ab902e4661/silnlp/nmt/hugging_face_config.py#L1243)):*
> ```Python
>             device=0,
>         )
>         pipeline.model = torch.compile(pipeline.model)
> ```

Should we pull the compilation step to outside the method and outside the for loop, so that we only compile once and pass the compiled model to `_create_pipeline_for_test_file`? I imagine we want to avoid needing to spend time compiling more than necessary, unless there's bugs you think could arise from only compiling once here.

benjaminking (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/987)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on ddaspit and mshannon-sil).

___
*[`silnlp/nmt/hugging_face_config.py` line 1243 at r1](https://reviewable.io/reviews/sillsdev/silnlp/987#-Or-qnfdBf0rEyiXfoD6:-Or06RsL6jiTFTmNLQHJ:b-dm3gzd) ([raw file](https://github.com/sillsdev/silnlp/blob/5356d22e84f45d6d24b1a4a311a522ab902e4661/silnlp/nmt/hugging_face_config.py#L1243)):*

Should we pull the compilation step to outside the method and outside the for loop, so that we only compile once and pass the compiled model to `_create_pipeline_for_test_file`? I imagine we want to avoid needing to spend time compiling more than necessary, unless there's bugs you think could arise from only compiling once here.
</blockquote></details>

If we could compile once, that it would be preferable.  The model needs to be uncompiled when it's passed to the pipeline, but then compiled before calling the pipeline.  So one option would be to compile the model right after it's created, and then pass both a compiled and uncompiled version to `_create_pipeline_for_test_file`.  I'll look into doing that.

benjaminking (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/987)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on ddaspit and mshannon-sil).

___
*[`silnlp/nmt/hugging_face_config.py` line 1243 at r1](https://reviewable.io/reviews/sillsdev/silnlp/987#-Or-qnfdBf0rEyiXfoD6:-OrPggsy3TEuVKg7t91M:b7iwe42) ([raw file](https://github.com/sillsdev/silnlp/blob/5356d22e84f45d6d24b1a4a311a522ab902e4661/silnlp/nmt/hugging_face_config.py#L1243)):*

If we could compile once, that it would be preferable.  The model needs to be uncompiled when it's passed to the pipeline, but then compiled before calling the pipeline.  So one option would be to compile the model right after it's created, and then pass both a compiled and uncompiled version to `_create_pipeline_for_test_file`.  I'll look into doing that.
</blockquote></details>

I have now implemented the change to only compile the model once and have verified that the test step still works correctly.

mshannon-sil (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/987#-:-OrPwgHu1B8u4rYLsVJJ:bnfp4nl)* status: 0 of 1 files reviewed, all discussions resolved (waiting on ddaspit).

PR #1011, opened 2026-04-27
Fix bug with repeated onboarding attempts

TaperChipmunk32:
The old way of checking if a request was already onboarded was inconsistent, leading to some projects being repeatedly onboarded. This should fix that issue by using list comprehension, rather than modifying the list while in a loop.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/1011)


benjaminking (approved):
> @benjaminking reviewed 1 file and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/1011)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on TaperChipmunk32).

PR #1020, opened 2026-04-29
Add a QA mode to the Auto Onboarding script for SF testing

TaperChipmunk32:
This PR adds a `--qa` arg that will use different environment variables to get the Onboarding Requests from SF QA instead of SF Live.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/1020)

### DIFF
--- a/scripts/automate_onboard_project.py
+++ b/scripts/automate_onboard_project.py
@@ -9,7 +9,7 @@ parser.add_argument("project", type=str, help="The name of the main Paratext pro
 parser.add_argument(
     "--dir",
     type=str,
-    help="The name of the directory in ONBOARDING_PATH to use as the source for onboarding.",
+    help="The name of the directory to use as the source for onboarding.",
 )
 parser.add_argument("--task-name", type=str, help="The name of the ClearML task to create for this onboarding process.")
 args = parser.parse_args()
@@ -29,7 +29,7 @@ try:
             "--security-opt apparmor=docker-apparmor",
             "--env CHECK_TRANSFERS=1",
             "--env SIL_NLP_DATA_PATH=/root/M",
-            f"-v {os.getenv('ONBOARDING_PATH')}/{args.dir}:/root/OnboardingProjects/{args.dir}",
+            f"-v {args.dir}:/root/OnboardingProjects/{args.dir}",
         ],
         docker_setup_bash_script=[
             "apt install -y python3-venv",
--- a/scripts/onboarding_requests.py
+++ b/scripts/onboarding_requests.py
@@ -12,54 +12,72 @@ from typing import Dict, Tuple
 import requests
 from clearml import Task
 
+LOGGER = logging.getLogger(__name__)
 
-def setup_environment(qa: bool):
-    global SF_AUTH_PWD, SF_CLIENT_ID, ONBOARDING_PATH, ONBOARDING_LOG_PATH, ONBOARDING_CLEANUP_PATH, ONBOARDING_REQUESTS_URL, PROJECTS_URL, SF_AUTH_URL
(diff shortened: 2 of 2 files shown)


benjaminking (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/1020)* status: all files reviewed, 1 unresolved discussion (waiting on TaperChipmunk32).

___
*[`scripts/onboarding_requests.py` line 16 at r1](https://reviewable.io/reviews/sillsdev/silnlp/1020#-OrOwVfh0CH31SsxL7EW:-OrOwVfh0CH31SsxL7EX:b4zm48x) ([raw file](https://github.com/sillsdev/silnlp/blob/fe70875cb588765db4e900365ee0a3bceb8093c9/scripts/onboarding_requests.py#L16)):*
> ```Python
> 
> 
> def setup_environment(qa: bool):
> ```

I'd think about creating a class to represent the onboarding environment, and all of these global variables can be class members.  You could then have two different class-level methods, e.g. `OnboardingEnvironment.createProductionEnvironment()` and `OnboardingEnvironment.createQAEnvironment()`.  

As we move toward creating more tests for SILNLP (and wrapping the onboarding code in tests would be a pretty high priority), it will make life easier to have things like this encapsulated in objects that you can inject, rather than in global variables.

TaperChipmunk32 (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/1020)* status: 1 of 2 files reviewed, 1 unresolved discussion (waiting on benjaminking).

___
*[`scripts/onboarding_requests.py` line 16 at r1](https://reviewable.io/reviews/sillsdev/silnlp/1020#-OrOwVfh0CH31SsxL7EW:-OrPFaeD6hTu35lkb2-0:b-896fix) ([raw file](https://github.com/sillsdev/silnlp/blob/fe70875cb588765db4e900365ee0a3bceb8093c9/scripts/onboarding_requests.py#L16)):*

I'd think about creating a class to represent the onboarding environment, and all of these global variables can be class members.  You could then have two different class-level methods, e.g. `OnboardingEnvironment.createProductionEnvironment()` and `OnboardingEnvironment.createQAEnvironment()`.  

As we move toward creating more tests for SILNLP (and wrapping the onboarding code in tests would be a pretty high priority), it will make life easier to have things like this encapsulated in objects that you can inject, rather than in global variables.
</blockquote></details>

Done.

benjaminking (approved):
:lgtm:

> @benjaminking reviewed 1 file and all commit messages, and made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/1020#-:-OrPSR--DTNtD647o1QU:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on TaperChipmunk32).

___
*[`scripts/onboarding_requests.py` line 19 at r2](https://reviewable.io/reviews/sillsdev/silnlp/1020#-OrPRGY1844VkADNHc5Z:-OrPRGY1844VkADNHc5_:bp0og8v) ([raw file](https://github.com/sillsdev/silnlp/blob/0016ec7fb2b55397f0da9e2813744568ca81bef8/scripts/onboarding_requests.py#L19)):*
> ```Python
> 
> class OnboardingEnvironment:
>     SF_AUTH_PWD: str = None
> ```

Just a note (and not something you need to do right now):  When we do start creating tests for onboarding, you'll probably want these to be instance-level variables instead of class-level variables, since it will still be possible for one test to inadvertently change one of these values for a different test.  And then you'll also want to pass the environment into your functions instead of it being global, but those can be tasks for a different day.

PR #1021, opened 2026-04-29
add support for chapter level usfm filtering and remarks

### LINKED ISSUE #836: Partial book drafts are difficult to import if the project has existing text for that book
When a partial book draft is created, the draft contains the intro/chapter/verse structure for the entire book, with draft text for only the specified chapters.  If one of these drafts is imported into a project as an update to an existing book with some existing verse text, the existing book will be overwritten with the new draft.  This would cause the loss of the existing text in the project.

An option is needed to allow only the specified chapters to be included in the partial book draft, with the unspecified chapters (and verses) excluded from the draft file.  This will allow the partial book draft to be imported into the Paratext project as an overlay for the existing text (i.e., it will only import the chapters included in the partial book draft, and leave the other chapters unchanged.)

mshannon-sil:
This PR addresses issue #836 and is dependent on changes in machine.py that are in this PR: https://github.com/sillsdev/machine.py/pull/285.

This passes the chapters argument into the `update_usfm` method from machine.py so that any usfm markers that are not the \id marker or in the requested chapters are omitted. It also updates remarks to be applied at a chapter level rather than at a book level.

A couple questions/notes:
- For books with a single chapter like Jude, are we okay always assuming that there will be a `\c 1` marker, or is it possible projects may omit it? Because the way the remarks algorithm now works for a book like Jude, it creates a remark for chapter 1, but the remark won't be applied if there is no `\c 1` marker.
- I didn't update the more manual version of updating usfm in the `translate_usfm` method, since the way silnlp is set up it doesn't get used when books/chapters are specified. Also, updating it to support chapters would include having to import more from machine.py, particularly `filter_tokens_by_chapter` that we didn't want to import if we didn't have to. If I should still add support for chapter level drafting to this manual version regardless, please let me know.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/1021)

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -33,8 +33,8 @@ torch = ">=1.10.0"
 
 [package.extras]
 deepspeed = ["deepspeed"]
-dev = ["bitsandbytes", "black (~=23.1)", "datasets", "diffusers", "evaluate", "hf-doc-builder (>=0.3.0)", "parameterized", "pytest (>=7.2.0,<=8.0.0)", "pytest-subtests", "pytest-xdist", "rich", "ruff (~=0.6.4)", "scikit-learn", "scipy", "timm", "torchdata (>=0.8.0)", "torchpippy (>=0.2.0)", "tqdm", "transformers"]
-quality = ["black (~=23.1)", "hf-doc-builder (>=0.3.0)", "ruff (~=0.6.4)"]
+dev = ["bitsandbytes", "black (>=23.1,<24.0)", "datasets", "diffusers", "evaluate", "hf-doc-builder (>=0.3.0)", "parameterized", "pytest (>=7.2.0,<=8.0.0)", "pytest-subtests", "pytest-xdist", "rich", "ruff (>=0.6.4,<0.7.0)", "scikit-learn", "scipy", "timm", "torchdata (>=0.8.0)", "torchpippy (>=0.2.0)", "tqdm", "transformers"]
+quality = ["black (>=23.1,<24.0)", "hf-doc-builder (>=0.3.0)", "ruff (>=0.6.4,<0.7.0)"]
 rich = ["rich"]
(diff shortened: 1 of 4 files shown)


Enkidu93 (approved):
Looks good!

I think we can assume that `\c 1`  always exists. I've never seen it omitted I don't think. And if it were omitted, I think it would cause other problems in our tools. Maybe you could ask the EITL team if they've seen this before if you'd like to double-check?

I'm fine with only updating `translate_book`, but that's more a question for Ben and Damien, I guess.

> @Enkidu93 reviewed 2 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/1021#-:-OsbHC1QAHQgF8MuSglH:batuw18)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on ddaspit).

ddaspit (approved):
:lgtm:

> @ddaspit reviewed 2 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/1021#-:-OsgfOtd31bSRztOPyDd:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on mshannon-sil).

benjaminking (approved):
I'm fine with not having this behavior apply to the `translate_usfm` path, since it doesn't include a way to pass chapters, as you said.
:lgtm:

> @benjaminking reviewed all commit messages and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/1021#-:-Ot-wavc8TS2_NJplQUx:b-jaufa4)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on mshannon-sil).

PR #1035, opened 2026-04-30
resize checkpoint embeddings if tokenizer vocab mismatch

### LINKED ISSUE #1033: cuda out of bounds when tokenizer vocab size != checkpoint embedding size
When running inference with a checkpoint whose saved embedding size differs from the current tokenizer's vocab size, a CUDA device-side assertion is triggered during beam search. This happens when running preprocess and test on a different source text that may have more unknown tokens than the original source/target texts the model was trained on.
```
../aten/src/ATen/native/cuda/IndexKernel.cu:92: operator(): block: [0,0,0], thread: [31,0,0] Assertion `-sizes[i] <= index && index < sizes[i] && "index out of bounds"` failed.
  0%|                                                           | 0/100 [00:38<?, ?ex/s]
Traceback (most recent call last):
  File "/root/silnlp/silnlp/nmt/experiment.py", line 317, in <module>
    main()
  File "/root/silnlp/silnlp/nmt/experiment.py", line 313, in main
    exp.run()
  File "/root/silnlp/silnlp/nmt/experiment.py", line 54, in run
    self.test()
  File "/root/silnlp/silnlp/nmt/experiment.py", line 77, in test
    test(
  File "/root/silnlp/silnlp/nmt/test.py", line 822, in test
    results[step] = test_checkpoint(
  File "/root/silnlp/silnlp/nmt/test.py", line 597, in test_checkpoint
    model.translate_test_files(
  File "/root/silnlp/silnlp/nmt/hugging_face_config.py", line 1275, in translate_test_files
    sentence_translation_groups: List[SentenceTranslationGroup] = list(
  File "/root/silnlp/silnlp/nmt/hugging_face_config.py", line 1338, in _translate_test_sentences
    for model_output_group in tqdm(
  File "/usr/local/lib/python3.10/dist-packages/tqdm/std.py", line 1181, in __iter__
    for obj in iterable:
  File "/root/silnlp/silnlp/nmt/hugging_face_config.py", line 1613, in _translate_sentences
    yield from self._translate_sentence_helper(
  File "/root/silnlp/silnlp/nmt/hugging_face_config.py", line 1722, in _translate_sentence_helper
    for translated_sentence in self._translate_with_beam_search(
  File "/root/silnlp/silnlp/nmt/hugging_face_config.py", line 1762, in _translate_with_beam_search
    translations = [[t] for t in translations]
  File "/root/silnlp/silnlp/nmt/hugging_face_config.py", line 1762, in <listcomp>
    translations = [[t] for t in translations]
  File "/usr/local/lib/python3.10/dist-packages/transformers/pipelines/pt_utils.py", line 124, in __next__
    item = next(self.iterator)
  File "/usr/local/lib/python3.10/dist-packages/transformers/pipelines/pt_utils.py", line 125, in __next__
    processed = self.infer(item, **self.params)
  File "/usr/local/lib/python3.10/dist-packages/transformers/pipelines/base.py", line 1209, in forward
    model_outputs = self._forward(model_inputs, **forward_params)
  File "/root/silnlp/silnlp/nmt/hugging_face_config.py", line 2042, in _forward
    output = self.model.generate(
  File "/usr/local/lib/python3.10/dist-packages/torch/utils/_contextlib.py", line 116, in decorate_context
    return func(*args, **kwargs)
  File "/usr/local/lib/python3.10/dist-packages/transformers/generation/utils.py", line 2246, in generate
    result = self._beam_search(
  File "/usr/local/lib/python3.10/dist-packages/transformers/generation/utils.py", line 3515, in _beam_search
    next_token_scores, next_tokens = torch.topk(
RuntimeError: CUDA error: device-side assert triggered
CUDA kernel errors might be asynchronously reported at some other API call, so the stacktrace below might be incorrect.
For debugging consider passing CUDA_LAUNCH_BLOCKING=1
Compile with `TORCH_USE_CUDA_DSA` to enable device-side assertions.
```

mshannon-sil:
This addresses #1033. If the loaded checkpoint vocab size does not match the tokenizer vocab size, the checkpoint's embedding are now resized. This allows for trying to inference from multiple source texts without needing to worry about an extra unknown character causing an exception. I did make the logger output a warning when this happens though, since it's not ideal to update the embedding for the model at inference time without then training those embeddings.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/1035)

### DIFF
--- a/silnlp/nmt/hugging_face_config.py
+++ b/silnlp/nmt/hugging_face_config.py
@@ -1856,8 +1856,6 @@ class HuggingFaceNMTModel(NMTModel):
             model: PreTrainedModel = self._pretrained_model_provider.create_model_for_inference(model_name)
         if self._config.infer.get("better_transformer"):
             model = model.to_bettertransformer()
-        if model_name == self._config.model and len(tokenizer) != model.get_input_embeddings().weight.size(dim=0):
-            model.resize_token_embeddings(len(tokenizer), pad_to_multiple_of=8 if self._mixed_precision else None)
         model, tokenizer = self._configure_model(model, tokenizer, src_lang, trg_lang)
 
         if model.generation_config is not None and (
@@ -1906,6 +1904,13 @@ class HuggingFaceNMTModel(NMTModel):
             forced_bos_token_id = tokenizer.convert_tokens_to_ids(trg_lang)
             if model.generation_config is not None:
                 model.generation_config.forced_bos_token_id = forced_bos_token_id
+
+        if len(tokenizer) != model.get_input_embeddings().weight.size(dim=0):
+            LOGGER.warning(
+                f"Tokenizer vocab size ({len(tokenizer)}) does not match model's input embedding vocab size "
+                f"({model.get_input_embeddings().weight.size(dim=0)}). Resizing model embeddings."
+            )
+            model.resize_token_embeddings(len(tokenizer), pad_to_multiple_of=8 if self._mixed_precision else None)
 
         return model, tokenizer
 


benjaminking (approved):
This solution looks good.  I do wonder however, whether this is a use case we should support?  In general, we have no guarantees that the tokens from two different tokenizers will be in any way compatible with one another.  And it seems like the size mismatch is an indicator that they likely will not be compatible.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/1035#-:-OrUTK9L25yHDY4-bktb:bdlmu5v)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on mshannon-sil).

mshannon-sil:
Hmm that is true, especially if the first tokenizer has multiple unknown tokens added, and the second one has those plus a few more, I don't think it's guaranteed they'll be added in the same order. The use case the EITL team had where they encountered this bug was being testing the same model across multiple different source texts. This should normally work already, the exception is when the different source texts have different amount of unknown tokens.

Do we want to throw an error here then instead? And just have the EITL train a new model if they want test results on a different source text where there's a different number of unknown characters? I imagine that's probably the more correct method rather than potentially introduce silent errors.

benjaminking:
Maybe we should plan to discuss that at our next SILNLP meeting.  It should be fine to merge now, and we can talk later about what use cases we want to support.

PR #1057, opened 2026-05-05
Refactor and fix bug with generating resource hashes

TaperChipmunk32:
This PR refactors when the resource hash is generated for a PT Resource. The old version contained a bug that would cause a crash with the `.resource_hash` not being found. It also fixes another bug where the `generate_resource_hash` would only generate a hash for the version on the bucket.

Now, it will always generate a hash and store it in the `.resource_hash` file for both the version being onboarded and the version on the bucket, if it exists and the `.resource_hash` file does not exist.

Here are the 3 different results of comparing the hashes:
- If the hashes are the same, the onboarding version will not be uploaded. If it is different,
- If the hashes are different, the bucket version will be renamed with a current datestamp and the onboarding version will be uploaded with no datestamp
- If the Resource does not exist on the bucket already, the onboarding version will be uploaded.

This rework/fix should be both more accurate and not require old resources to always be reuploaded because they do not have a `.resource_hash` file.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/1057)

### DIFF
--- a/scripts/onboarding_requests.py
+++ b/scripts/onboarding_requests.py
@@ -190,17 +190,14 @@ def append_datestamp(project_name: str) -> str:
     return f"{project_name}_{datestamp}"
 
 
-def rename_project(project_name: str, datestamp: bool, copy_from: Path | None) -> str:
-    is_resource = False
+def rename_project(project_name: str, datestamp: bool) -> str:
+    resource = False
     if project_name.endswith("_Resource"):
-        is_resource = True
-        resource_hash_path = copy_from / Path(f"{project_name}/.resource_hash") if copy_from else None
-        if resource_hash_path and not resource_hash_path.exists():
-            resource_hash_path.touch()
+        resource = True
         project_name = project_name.replace("_Resource", "")
     if "-" in project_name:
         project_name = project_name.replace("-", "_")
-    if datestamp and not is_resource:
+    if datestamp and not resource:
         project_name = append_datestamp(project_name)
     return project_name
 
@@ -237,9 +234,7 @@ def process_request(request):
             request["id"],
             f"This request is being automatically onboarded.\nClearML task: {task_name}.\nLink: {task.get_output_log_web_page()}",
         )
-        adjusted_name = rename_project(
-            main_project_name, True, Path(f"{OnboardingEnvironment.ONBOARDING_PATH}/{main_project_name}_Request")
-        )
+        adjusted_name = rename_project(main_project_name, True)
         add_comment(
             request["id"],
             f"Results will be stored in {OnboardingEnvironment.ONBOARDING_REQUESTS_BUCKET_DIR}/{adjusted_name}",
--- a/silnlp/common/onboard_project.py
+++ b/silnlp/common/onboard_project.py
@@ -283,7 +283,7 @@ class OnboardingProject:
 
         self.check_for_project_errors()
 
-        self.project_name = rename_project(project_name=self.project_name, datestamp=datestamp, copy_from=copy_from)
+        self.project_name = self.rename_project(project_name=self.project_name, datestamp=datestamp)
 
         self.local_project_path = copy_project(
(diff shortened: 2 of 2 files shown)


benjaminking (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/1057)* status: all files reviewed, 1 unresolved discussion (waiting on TaperChipmunk32).

___
*[`silnlp/common/onboard_project.py` line 351 at r1](https://reviewable.io/reviews/sillsdev/silnlp/1057#-Ory9Z8z4ly9P5cKGCZ_:-Ory9Z8z4ly9P5cKGCZa:b-wpizgn) ([raw file](https://github.com/sillsdev/silnlp/blob/8c84881164d2740bf26398772228ae2d1606964b/silnlp/common/onboard_project.py#L351)):*
> ```Python
>     def rename_project(self, project_name: str, datestamp: bool) -> str:
>         self.resource = False
>         if project_name.endswith("_Resource"):
> ```

I think I'm missing something, but wouldn't it be possible for projects that get renamed via this method to fail to get a `.resource_hash` file created?

TaperChipmunk32:
*[`silnlp/common/onboard_project.py` line 351 at r1](https://reviewable.io/reviews/sillsdev/silnlp/1057#-Ory9Z8z4ly9P5cKGCZ_:-OryBKZeCQmpX9EbsSCX:b7luyo6) ([raw file](https://github.com/sillsdev/silnlp/blob/8c84881164d2740bf26398772228ae2d1606964b/silnlp/common/onboard_project.py#L351)):*

I think I'm missing something, but wouldn't it be possible for projects that get renamed via this method to fail to get a `.resource_hash` file created?
</blockquote></details>

The `.resource_hash` file is only created if the project name ends with `_Resource`, which is appended to the name of all Resources in `onboarding_requests.py`. The `.resource_hash` file is only needed for Resources, not for all projects.

TaperChipmunk32:
*[`silnlp/common/onboard_project.py` line 351 at r1](https://reviewable.io/reviews/sillsdev/silnlp/1057#-Ory9Z8z4ly9P5cKGCZ_:-OryCDc90XIjjWwDL9ZR:bm3huq4) ([raw file](https://github.com/sillsdev/silnlp/blob/8c84881164d2740bf26398772228ae2d1606964b/silnlp/common/onboard_project.py#L351)):*

The `.resource_hash` file would only not be created if the project name does not end with `_Resource`, which is appended to the name of all Resources in `onboarding_requests.py`. The `.resource_hash` file is only needed for Resources, not for all projects.
</blockquote></details>

This does make me think of an idea. We could extend this logic to all projects, including a hash file to check their contents. Because it is likely that we have copies of projects with different datestamps that are actually identical. The same way we now treat Resources could be the way we treat all projects, but we would need to discuss this with the EITL team.

benjaminking (approved):
:lgtm:

> @benjaminking made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/1057#-:-OryGHl57xnWFMjX_7CT:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on TaperChipmunk32).

___
*[`silnlp/common/onboard_project.py` line 351 at r1](https://reviewable.io/reviews/sillsdev/silnlp/1057#-Ory9Z8z4ly9P5cKGCZ_:-OryGGEPD8XNMr6aRzAt:b-9cs2bb) ([raw file](https://github.com/sillsdev/silnlp/blob/8c84881164d2740bf26398772228ae2d1606964b/silnlp/common/onboard_project.py#L351)):*

This does make me think of an idea. We could extend this logic to all projects, including a hash file to check their contents. Because it is likely that we have copies of projects with different datestamps that are actually identical. The same way we now treat Resources could be the way we treat all projects, but we would need to discuss this with the EITL team.
</blockquote></details>

Ok, I think I understand now.

PR #1062, opened 2026-05-08
`SilNlpEnv` refactoring

benjaminking:
This PR includes two major changes to `SilNlpEnv`:

1. It is now immutable, with factory methods to initialize certain paths to custom values
2. It is no longer used as a global variable, but is instead injected (passed) into each context where it is used

I should mention that this refactoring is far from the final word on how the environment should be treated.  It's a step in the right direction, but in most places where `SilNlpEnv` is used, there is further refactoring to do.  It is now frequently an argument passed to various functions, and in almost all of these cases, it would be more appropriate to wrap these functions in classes, so that long-lived context information, like the environment, can be provided during object creation, and the request-scoped information can remain as the method parameters.  Furthermore, many places in the code should probably not be interacting directly with the environment object, but with a higher-level abstraction (something like `ExperimentFileManager` or `ParatextProject`).

@TaperChipmunk32, can you take a close look at the onboarding scripts?  I had to make several changes to those scripts, but I don't know the right way to test them, so those changes have not been tested.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/1062)


TaperChipmunk32 (changes requested):
> @TaperChipmunk32 reviewed 66 files and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/1062)* status: all files reviewed (commit messages unreviewed), 1 unresolved discussion (waiting on benjaminking).

___
*[`silnlp/common/environment.py` line 79 at r1](https://reviewable.io/reviews/sillsdev/silnlp/1062#-OsMdeOSDWv0YTti9u48:-OsMdeOSDWv0YTti9u49:b-it261v) ([raw file](https://github.com/sillsdev/silnlp/blob/ad836aab019848b2c19a77f0693f3a7bdc06b6e8/silnlp/common/environment.py#L79)):*
> ```Python
>             return self._data_dir / custom_dir
> 
>         if env_var_dir is not None:
> ```

Both top level if statements seem to contradict the comment at the top of this function. They both give precedence to environment variables when there is an explicit path given. If I understand this correctly. both ifs should be `if custom_dir is None` and `if env_var_dir is None`.

TaperChipmunk32:
@benjaminking I looked through `onboard_project` and could not find any issues with it.

benjaminking (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/1062)* status: 65 of 66 files reviewed, 1 unresolved discussion (waiting on TaperChipmunk32).

___
*[`silnlp/common/environment.py` line 79 at r1](https://reviewable.io/reviews/sillsdev/silnlp/1062#-OsMdeOSDWv0YTti9u48:-OsMsb6yCQ7Ug2Vi-hVK:btkwpq7) ([raw file](https://github.com/sillsdev/silnlp/blob/ad836aab019848b2c19a77f0693f3a7bdc06b6e8/silnlp/common/environment.py#L79)):*

Both top level if statements seem to contradict the comment at the top of this function. They both give precedence to environment variables when there is an explicit path given. If I understand this correctly. both ifs should be `if custom_dir is None` and `if env_var_dir is None`.
</blockquote></details>

I'm not sure I'm seeing the issue that you're bringing up.  The logic is supposed to be: 1. If there's a custom_dir passed, use that, 2. If not and there's an environment variable dir, use that, and 3. otherwise use the default_subdir.  (I also updated the comment to try to make that more clear)  Is there something else I'm missing?

TaperChipmunk32 (approved):
:lgtm:

> @TaperChipmunk32 made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/1062#-:-OsNVUyS3kZS2j4DnsD4:bnfp4nl)* status: 65 of 66 files reviewed, 1 unresolved discussion (waiting on benjaminking).

___
*[`silnlp/common/environment.py` line 79 at r1](https://reviewable.io/reviews/sillsdev/silnlp/1062#-OsMdeOSDWv0YTti9u48:-OsNV-LFEw9lqPixHxPe:b-4fzxam) ([raw file](https://github.com/sillsdev/silnlp/blob/ad836aab019848b2c19a77f0693f3a7bdc06b6e8/silnlp/common/environment.py#L79)):*

I'm not sure I'm seeing the issue that you're bringing up.  The logic is supposed to be: 1. If there's a custom_dir passed, use that, 2. If not and there's an environment variable dir, use that, and 3. otherwise use the default_subdir.  (I also updated the comment to try to make that more clear)  Is there something else I'm missing?
</blockquote></details>

Oh I see now, I was misinterpreting the second if statement. Thanks for the clarification.

benjaminking:
@copilot resolve the merge conflicts in this pull request, making sure that the new design for the environment is respected, with an instance of `SilNlpEnv` being injected wherever possible, added to the end of the arguments list, and only having a default value if it occurs after other arguments with default values.

PR #1085, opened 2026-06-01
Fix line number miscounting

TaperChipmunk32:
This fixes an issue with miscounting the number of lines in an extract file.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/1085)

### DIFF
--- a/silnlp/common/extract_corpora.py
+++ b/silnlp/common/extract_corpora.py
@@ -106,9 +106,9 @@ def extract_corpora(
                 LOGGER.info(
                     f"The number of completed verses is {verse_count} (out of the expected {expected_verse_count})."
                 )
-            if line_count > expected_verse_count:
-                LOGGER.warning(
-                    f"The number of lines in the corpus file is {line_count}, which is greater than the expected {expected_verse_count}."
+            if line_count != expected_verse_count:
+                LOGGER.error(
+                    f"The number of lines in the corpus file is {line_count}, which is not equal to the expected {expected_verse_count}."
                 )
             terms_count = extract_term_renderings(
                 project_dir, corpus_filename, environment.mt_terms_dir, extract_surface_forms, environment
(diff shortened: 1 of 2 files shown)


benjaminking (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/1085#-:-Ou2r0nBE3VBSFsI7Hzq:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on TaperChipmunk32).

mmartin9684-sil:
@TaperChipmunk32 @benjaminking - I don't think this change is getting at the real issue that we want to have flagged as an error.  It looks like it's doing useful filtering as it's counting the lines in the verse extract file that have actual text on them, so that we can see how many verses in the project have been translated. 
However, the issue that we want to have flagged as an error is when the actual # of lines in the file is not the expected number of lines (41,899).  That's a hard stop error - if the line count is wrong, then the extract got out of alignment somewhere, and the extract file is no good to use.  We need to know that.
As far as counting the number of lines in the extract with verse translations, it looks like the latest logic excludes blank lines, lines with "...", and lines with "<range>" as the text.  The first two checks make sense, but excluding the "<range>" lines is probably not correct.  If the project has a verse range (e.g., "\v 1-2"), that should actually be counted as 2 translated verses.   This code would count it as 1 translated verse because it would exclude the "<range>" line.

TaperChipmunk32:
@mmartin9684-sil I believe this does fix the issue. There are two different variables in this section that count the lines differently. `line_count` counts the number of lines in the extract file, regardless of content. The bug was that `line_count` was skipping `<range>` lines, causing the # of lines to be reported as less than the actual number of lines in the file. This PR fixes that issue. 

`verse_counts` is the one that has all of the filtering to count lines with actual verse content. I'll work on changing this to properly count verse ranges in a separate PR, rather than excluding all `<range>` lines and the following empty lines.

PR #1087, opened 2026-06-02
Compute transition scores on CPU

benjaminking:
This PR moves the computation of transition scores in `SilTranslationPipeline` to the CPU.  The scores form a large tuple whose size is `num_generated_tokens × batch_size × num_beams × vocab_size` that can occupy a lot of the GPU's RAM.  The tradeoff is that inference now takes 1.5-2x longer, since we're using the CPU.

We should discuss further whether this is a good tradeoff to make.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/1087)

### DIFF
--- a/silnlp/nmt/hugging_face_config.py
+++ b/silnlp/nmt/hugging_face_config.py
@@ -1885,14 +1885,15 @@ class SilTranslationPipeline(TranslationPipeline):
 
     @staticmethod
     def _move_scores_and_beam_indices_to_cpu_for_long_sequences(
-        output_scores: Tuple[Tensor, ...], beam_indices: Optional[Tensor]
-    ) -> Tuple[Tuple[Tensor, ...], Optional[Tensor]]:
+        output_ids: Tensor, output_scores: Tuple[Tensor, ...], beam_indices: Optional[Tensor]
+    ) -> Tuple[Tensor, Tuple[Tensor, ...], Optional[Tensor]]:
         if max(len(score) for score in output_scores) > 200:
             return (
+                output_ids.to("cpu"),
                 tuple(score.to("cpu") for score in output_scores),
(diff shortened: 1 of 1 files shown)


benjaminking (commented):
One upside to the performance hit that I just realized: the EITL team has already been running inference with a batch size of 8 instead of 16, so computing transition scores on the CPU with a batch size of 16 will have comparable performance to computing transition scores on the GPU with a batch size of 8.  So, this fix is at least no slower than the status quo.

Also, it looks like we compute transition scores whether or not we're calculating confidence.  We may want to change that so to avoid an unnecessary performance hit calculating something that we're just throwing away.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/1087#-:-Ou8zhyDC9DWwjGiz6Gp:b-qc9jud)* status: 0 of 1 files reviewed, all discussions resolved.

benjaminking (commented):
I have updated this branch and changed the approach from always computing transition scores on the CPU to instead attempting to perform the calculations on the GPU and failing over to the CPU on any exception.  I have verified that this allows an experiment that consistently produces OOM errors to complete with no issues.  This also results in an order-of-magnitude speed up compared to the CPU-only approach.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/1087#-:-Ox16kC60UhNG-CixVKB:b-f149wi)* status: 0 of 1 files reviewed, all discussions resolved.

mshannon-sil (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/1087#-:-OxggE5Z8uUqt_Ez-KNO:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on benjaminking).

Issue #1079, opened 2026-05-28
Quotation denormalization remarks need to be updated for idg


mshannon-sil:
When quotation denormalization is selected, this error is ocurring:

```
2026-05-28 13:26:24,334 - silnlp.nmt.translate - ERROR - Was not able to translate JON.
Traceback (most recent call last):
  File "/root/.clearml/venvs-builds/3.10/task_repository/silnlp.git/silnlp/nmt/translate.py", line 159, in translate_books
    translator.translate_book(
  File "/root/.clearml/venvs-builds/3.10/task_repository/silnlp.git/silnlp/common/translator.py", line 352, in translate_book
    self.translate_usfm(
  File "/root/.clearml/venvs-builds/3.10/task_repository/silnlp.git/silnlp/common/translator.py", line 517, in translate_usfm
    usfm_out = quotation_denormalization_postprocessor.postprocess_usfm(usfm_out)
  File "/root/.clearml/venvs-builds/3.10/task_repository/silnlp.git/silnlp/common/postprocesser.py", line 297, in postprocess_usfm
    return handler.get_usfm()
  File "/usr/local/lib/python3.10/dist-packages/machine/corpora/update_usfm_parser_handler.py", line 344, in get_usfm
    for chapter_num, remark in self._remarks:
ValueError: too many values to unpack (expected 2)
```

We need to adjust this separate spot in silnlp that handles remarks to work with the latest changes to machine.py.

mshannon-sil:
There's an issue with the `DenormalizeQuotationMarksPostprocessor.get_chapters()` in machine.py. It's mislabeling the chapter numbers (each chapter is given the next chapter marker's number), so `find_best_chapter_strategies()` produces bad output. I'll do a workaround for now, but this will need to be fixed in machine.py.

Issue #1078, opened 2026-05-28
Update per-chapter remarks in silnlp to be consistent with SF/Serval remarks


mmartin9684-sil:
The per-chapter remarks inserted into the silnlp drafts should be consistent with the remarks inserted by SF/Serval, so that the user gets consistent drafts when working with EITL and with SF self-service.

Enkidu93:
The silnlp remark is specified [here](https://github.com/sillsdev/silnlp/blob/99482a891098fd4813e97875ab0bbff5bcbbbaf9/silnlp/common/translator.py#L448-L452) and the updates to the Serval remark are  [here](https://github.com/sillsdev/serval/pull/934).

PR #1090, opened 2026-06-05
Fix quotation denormalization, adjust remarks to match SF

mshannon-sil:
This PR addresses issue #1079 and #1078. It gets quotation denormalization working again in the `experiment.py --translate` step, as well as adjust the remarks to match what SF is producing for incremental draft generation.

I tested it and the remarks now look like the following:
```
\rem This draft of ACT 11 was generated by AI from project NIV11 on 2026-06-05 17:15:25Z. It should be reviewed carefully for errors before use. Paragraph break positions were preserved. Quotation marks have been adjusted automatically to match the rest of the project.
```
or alternatively for stripping paragraph markers and not doing quote denormalization:
```
\rem This draft of ACT 11 was generated by AI from project NIV11 on 2026-06-05 17:15:27Z. It should be reviewed carefully for errors before use. Paragraph break positions were removed.
```

This is not a complete fix for #1079, as there's changes to `QuotationMarkDenormalizationFirstPass.get_chapters()` in machine.py that need to happen so that it works properly and we can then simplify the silnlp code a bit.

This also might not be a complete fix for #1078, since if we want to create remarks for placing paragraph marker positions at the end, that would involve adding another enum value to UpdateUsfmMarkerBehavior in machine.py. I could also envision further small adjustments happening to the SF remarks at the moment based on the latest email discussion I saw, so I want to get final confirmation from them. But this change gets us 95% of the way there at the very least.

Since the EITL team is blocked on creating drafts with denormalized quotations at the moment, we'll get this PR in, and then have a further PR to wrap up loose ends.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/1090)


benjaminking (changes requested):
> @benjaminking reviewed 2 files and all commit messages, and made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/1090)* status: all files reviewed, 2 unresolved discussions (waiting on mshannon-sil).

___
*[`silnlp/common/postprocesser.py` line 246 at r1](https://reviewable.io/reviews/sillsdev/silnlp/1090#-OuOETlm-sSjRafN-jPt:-OuOETlnDx8WoO1eiZ4K:bh4clbt) ([raw file](https://github.com/sillsdev/silnlp/blob/cd8bd163902d7a9a050f33cce98433220d6d0386/silnlp/common/postprocesser.py#L246)):*
> ```Python
> 
>     @staticmethod
>     def _get_chapter_number(chapter) -> Optional[int]:
> ```

Could you add a type hint here?
___
*[`silnlp/common/postprocesser.py` line 269 at r1](https://reviewable.io/reviews/sillsdev/silnlp/1090#-OuOFZ3pEMwtE0oUld2r:-OuOFZ3q-XV_DuQAd8as:biv6lcd) ([raw file](https://github.com/sillsdev/silnlp/blob/cd8bd163902d7a9a050f33cce98433220d6d0386/silnlp/common/postprocesser.py#L269)):*
> ```Python
>     ) -> List[Tuple[int, str]]:
>         if not remarks:
>             return []
> ```

Is this a case we ever expect to happen?  Because it seems like we'd still want to include the quotation denormalization remarks even if the other remarks weren't generated for some reason.

mshannon-sil:
*[`silnlp/common/postprocesser.py` line 269 at r1](https://reviewable.io/reviews/sillsdev/silnlp/1090#-OuOFZ3pEMwtE0oUld2r:-OuOUdzFCVFBiaWA70Jt:bpw4yae) ([raw file](https://github.com/sillsdev/silnlp/blob/cd8bd163902d7a9a050f33cce98433220d6d0386/silnlp/common/postprocesser.py#L269)):*

Is this a case we ever expect to happen?  Because it seems like we'd still want to include the quotation denormalization remarks even if the other remarks weren't generated for some reason.
</blockquote></details>

This shouldn't ever happen. It's only optional because remarks is marked as optional in postprocess_usfm, but we always pass it in. Also if we tried to get this method to work if remarks is empty, it would fail because we wouldn't know which chapters from chapter strategies to use, so I'd need to do a bit more refactoring to support that use case.

mshannon-sil:
*[`silnlp/common/postprocesser.py` line 246 at r1](https://reviewable.io/reviews/sillsdev/silnlp/1090#-OuOETlm-sSjRafN-jPt:-OuOUFPq2ex6QpL6LCQW:b-896fix) ([raw file](https://github.com/sillsdev/silnlp/blob/cd8bd163902d7a9a050f33cce98433220d6d0386/silnlp/common/postprocesser.py#L246)):*

Could you add a type hint here?
</blockquote></details>

Done.

mshannon-sil:
I also see that postprocess.py does not work correctly. I spent the last few hours trying to get a fix in for that along with this, but it's way more involved to adjust and verify/test than I expected so I'll fix that separately.

mshannon-sil:
To clarify I mean when you run postprocess.py as the main script.

benjaminking (approved):
Ok, it sounds like the EITL doesn't often use the postprocessing scripts on their own, so it should be fine to address these cases separately.
:lgtm:

> @benjaminking made 1 comment and resolved 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/1090#-:-OuOgUQ3D1J8vii6FNQA:b3r4miw)* status: 1 of 2 files reviewed, all discussions resolved (waiting on mshannon-sil).

PR #1097, opened 2026-06-17
Add onboarding_report.csv generation

TaperChipmunk32:
This PR adds a `onboarding_report.csv` file that is generated at the end of an Onboarding Request with basic information on each of the projects all compiled in one file, rather than being spread across the command line output and various other output files.

- Added a `OnboardingReport` class that is used by each `OnboardingProject` to store and format the information needed for the csv file
- Added the `generate_report` and `create_report` functions to `OnboardingProject` and `OnboardingRequest` respectively to create the csv file using the values from the `OnboardingReport`for each project in the Request
- Added additional return values to `extract_corpora` and `check_versification` to pass along `detected_versification, `versification_error_count`, and `terms_count` to the `OnboardingProject` for the Onboarding Report.

This PR also addresses two small bugs:
- Added error catching when the main project fails to upload/Settings file is invalid
- Correcting a typo from `corpus_stats.csv` to `corpus-stats.csv`

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/1097)

### DIFF
--- a/silnlp/common/extract_corpora.py
+++ b/silnlp/common/extract_corpora.py
@@ -1,7 +1,7 @@
 import argparse
 import logging
 from pathlib import Path
-from typing import Optional, Set
+from typing import List, Optional, Set, Tuple
 
 from ..common.corpus import count_lines
 from ..common.environment import SilNlpEnv
@@ -76,7 +76,7 @@ def extract_corpora(
     parent_project: Optional[str] = None,
     versification_error_output_path: Optional[str] = None,
     environment: SilNlpEnv = SilNlpEnv.create_standard_environment(),
-) -> Path | None:
+) -> Tuple[Path | None, List[int], int, int]:
     # Process the projects that have data and tell the user.
     if len(projects) > 0:
         expected_verse_count = count_lines(environment.assets_dir / "vref.txt")
@@ -85,11 +85,17 @@ def extract_corpora(
         for project in projects:
             LOGGER.info(f"Extracting {project}...")
             project_dir = environment.get_paratext_project_dir(project)
-            parent_project_dir = get_parent_project_dir(project_dir, environment)
+
+            if parent_project is None:
+                parent_project_dir = get_parent_project_dir(project_dir, environment)
+            else:
+                parent_project_dir = environment.get_paratext_project_dir(parent_project)
             if parent_project_dir is not None:
                 LOGGER.info(f"Identified parent project {parent_project_dir.name}.")
 
-            check_versification(project_dir, parent_project_dir, versification_error_output_path, environment)
+            matching, detected_versification, versification_error_count = check_versification(
+                project_dir, parent_project_dir, versification_error_output_path, environment
+            )
             corpus_filename, verse_count, line_count = extract_project(
                 project_dir,
                 environment.mt_scripture_dir,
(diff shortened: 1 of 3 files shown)


TaperChipmunk32:
Switching this to a draft to wait for the next machine release to add `NormalizationForm` and `Language` to the report.

benjaminking (commented):
Sorry it took so long to get to this.

> @benjaminking reviewed 3 files and all commit messages, and made 5 comments.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/1097#-:-OwJyjtz4JOVYPGZqoC7:b-uxyesi)* status: all files reviewed, 4 unresolved discussions.

___
*[`silnlp/common/extract_corpora.py` line 79 at r2](https://reviewable.io/reviews/sillsdev/silnlp/1097#-OwJrzZYA0Y3YH5cPwIa:-OwJrzZYA0Y3YH5cPwIb:bhqae5r) ([raw file](https://github.com/sillsdev/silnlp/blob/a49056510257deb4a1b783ebd50380f5c87f7c9d/silnlp/common/extract_corpora.py#L79)):*
> ```Python
>     versification_error_output_path: Optional[str] = None,
>     environment: SilNlpEnv = SilNlpEnv.create_standard_environment(),
> ) -> Tuple[Path | None, List[int], int, int]:
> ```

When you have a return type that gets this complex, it's usually good to think about creating a class to represent the return value.  The class gives you some semantics about what the different parts mean, which helps to reduce programming errors.  For example, it would be really easy to accidentally mix up the terms count and the versification error count, but less easy with a class.
___
*[`silnlp/common/extract_corpora.py` line 96 at r2](https://reviewable.io/reviews/sillsdev/silnlp/1097#-OwJvgrk8xzvYSpBMSUW:-OwJvgrk8xzvYSpBMSUX:b9jae69) ([raw file](https://github.com/sillsdev/silnlp/blob/a49056510257deb4a1b783ebd50380f5c87f7c9d/silnlp/common/extract_corpora.py#L96)):*
> ```Python
>                 LOGGER.info(f"Identified parent project {parent_project_dir.name}.")
> 
>             matching, detected_versification, versification_error_count = check_versification(
> ```

It would similarly be good to have a class the represents the output of `check_versification`.
___
*[`silnlp/common/onboard_project.py` line 431 at r2](https://reviewable.io/reviews/sillsdev/silnlp/1097#-OwJxCdW4R1CSt5U0iWB:-OwJxCdW4R1CSt5U0iWC:b-33lic6) ([raw file](https://github.com/sillsdev/silnlp/blob/a49056510257deb4a1b783ebd50380f5c87f7c9d/silnlp/common/onboard_project.py#L431)):*
> ```Python
> 
>     def generate_report(self) -> dict:
>         self.report.name_on_bucket = self.project_name
> ```

Are you familiar with the "Builder" pattern?  It's a software design pattern used for creating a complex object (like `OnboardingReport`), and I think it would be a good fit for what you're doing here.  It would let you encapsulate the logic for building the report inside of the `OnboardingReport` class.  Accessing class members outside of the class is always a risk for bugs as the software evolves.
___
*[`silnlp/common/onboard_project.py` line 493 at r2](https://reviewable.io/reviews/sillsdev/silnlp/1097#-OwJySFM9eFAFtmma-NR:-OwJySFM9eFAFtmma-NS:b-k6wkty) ([raw file](https://github.com/sillsdev/silnlp/blob/a49056510257deb4a1b783ebd50380f5c87f7c9d/silnlp/common/onboard_project.py#L493)):*
> ```Python
>         self.main_project: OnboardingProject = OnboardingProject(
>             project_name=main_project_name,
>             project_type="Request Project",
> ```

The project type might be a good use case for an enum, since there's a small set of possible values it could take.

TaperChipmunk32:
@benjaminking I forgot to mention this, but [PR](https://github.com/sillsdev/silnlp/pull/1105) is a newer version of this PR.

PR #1103, opened 2026-06-26
Implement new linear regression method

benjaminking:
This PR includes the new method for performing linear regression on chrF3 and confidence scores in order to project chrF3 scores for quality estimation.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/1103)

### DIFF
--- a/silnlp/common/linear_regression.py
+++ b/silnlp/common/linear_regression.py
@@ -58,7 +58,7 @@ def perform_weighted_linear_regression(
     point_weighting_scheme: PointWeightingScheme = UniformPointWeightingScheme(),
 ) -> LinearRegressionResult:
     weights = point_weighting_scheme.weight_points(x, y)
-    sigmas = [1 / np.sqrt(w) if w > 0 else 0 for w in weights]
+    sigmas = [1 / np.sqrt(w) if w > 0 else np.inf for w in weights]
     popt, _ = curve_fit(linear_model, x, y, sigma=sigmas, absolute_sigma=True)
     slope, intercept = popt
     slope = round(slope, 4)
--- a/silnlp/common/linear_regression.py
+++ b/silnlp/common/linear_regression.py
@@ -1,0 +1,122 @@
+import json
+from dataclasses import dataclass
+from typing import List, Tuple
+
+import numpy as np
+from scipy.optimize import curve_fit
+from scipy.stats import linregress
+
+SCHEMA_VERSION = "0.1"
+DEFAULT_NUM_BUCKETS = 100
+
+
+@dataclass
+class LinearRegressionResult:
+    version: str
+    slope: float
+    intercept: float
+
+    def toJSON(self) -> str:
+        return json.dumps({"version": self.version, "slope": self.slope, "intercept": self.intercept}, indent=2)
+
+
+class PointWeightingScheme:
+    def weight_points(self, _x: List[float], _y: List[float]) -> List[float]: ...
+
+
+class UniformPointWeightingScheme(PointWeightingScheme):
+    def weight_points(self, x: List[float], y: List[float], **kwargs) -> List[float]:
+        return [1.0] * len(x)
+
+
+class InverseDensityPointWeightingScheme(PointWeightingScheme):
+    def weight_points(self, _x: List[float], y: List[float], num_buckets: int = DEFAULT_NUM_BUCKETS) -> List[float]:
+        density, _ = np.histogram(y, bins=num_buckets, range=(0, 100), density=True)
+        bin_edges = np.linspace(0, 100, num=num_buckets + 1)
+        weights = []
+        for y_instance in y:
+            bin_index = np.searchsorted(bin_edges, y_instance, side="right") - 1
+            bin_index = min(max(bin_index, 0), len(density) - 1)
+            weights.append(1 / (density[bin_index] * len(y) + 1e-6))
+        return weights
+
+
+def perform_simple_linear_regression(x: List[float], y: List[float]) -> LinearRegressionResult:
+    slope, intercept = linregress(x, y)[:2]
+    slope = round(slope, 4)
+    intercept = round(intercept, 4)
+    return LinearRegressionResult(version=SCHEMA_VERSION, slope=slope, intercept=intercept)
+
+
+def linear_model(x, slope, intercept):
+    return slope * x + intercept
+
+
+def perform_weighted_linear_regression(
+    x: List[float],
+    y: List[float],
+    point_weighting_scheme: PointWeightingScheme = UniformPointWeightingScheme(),
+) -> LinearRegressionResult:
+    weights = point_weighting_scheme.weight_points(x, y)
+    sigmas = [1 / np.sqrt(w) if w > 0 else 0 for w in weights]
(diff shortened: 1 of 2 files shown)


mshannon-sil (commented):
Could you also update the thresholds as part of this PR? I'm not sure what the usability values are, but the new projected chrF3 thresholds are 44.5 and 53.

> @mshannon-sil reviewed 2 files and all commit messages, and made 3 comments.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/1103#-:-Ow4Hi-1BP3JR8kwj8PU:bqjssvt)* status: all files reviewed, 2 unresolved discussions (waiting on benjaminking).

___
*[`silnlp/common/linear_regression.py` line 61 at r1](https://reviewable.io/reviews/sillsdev/silnlp/1103#-Ow4CdsS84a224B2CL7t:-Ow4CdsT7yTmzWWhfcAN:b-b59r96) ([raw file](https://github.com/sillsdev/silnlp/blob/ecaf42829237ddf83db2906711471aa9465fe481/silnlp/common/linear_regression.py#L61)):*
> ```Python
> ) -> LinearRegressionResult:
>     weights = point_weighting_scheme.weight_points(x, y)
>     sigmas = [1 / np.sqrt(w) if w > 0 else 0 for w in weights]
> ```

I think the else condition probably shouldn't be 0, since doesn't 0 sigma mean high confidence? Setting sigma to 0 also might result in divide by zero errors later on.
___
*[`silnlp/common/linear_regression.py` line 79 at r1](https://reviewable.io/reviews/sillsdev/silnlp/1103#-Ow4E7Qa6xrEruCsdw5O:-Ow4E7Qa6xrEruCsdw5P:bqbvy5p) ([raw file](https://github.com/sillsdev/silnlp/blob/ecaf42829237ddf83db2906711471aa9465fe481/silnlp/common/linear_regression.py#L79)):*
> ```Python
>         sampled_y = []
>         for _ in range(num_samples):
>             indices = np.random.choice(n, size=n, replace=True)
> ```

There's already an outer loop `for _ in range(100):` in `perform_enhanced_linear_regression`, so it seems like one of either the outer or inner loop are unnecessary for bootstrap sampling unless I'm missing something.

benjaminking (commented):
I have updated the usability parameters to match the new thresholds.

> @benjaminking made 3 comments.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/1103#-:-OwOt05wAv6gaItwEzqS:bnm84q1)* status: all files reviewed, 2 unresolved discussions (waiting on mshannon-sil).

___
*[`silnlp/common/linear_regression.py` line 61 at r1](https://reviewable.io/reviews/sillsdev/silnlp/1103#-Ow4CdsS84a224B2CL7t:-OwOoi5P8mQDV7smtKEN:by1n77b) ([raw file](https://github.com/sillsdev/silnlp/blob/ecaf42829237ddf83db2906711471aa9465fe481/silnlp/common/linear_regression.py#L61)):*

I think the else condition probably shouldn't be 0, since doesn't 0 sigma mean high confidence? Setting sigma to 0 also might result in divide by zero errors later on.
</blockquote></details>

Yes, good catch.  This was backwards.
___
*[`silnlp/common/linear_regression.py` line 79 at r1](https://reviewable.io/reviews/sillsdev/silnlp/1103#-Ow4E7Qa6xrEruCsdw5O:-OwOooZeBtW_CMaZLy39:bn1kydf) ([raw file](https://github.com/sillsdev/silnlp/blob/ecaf42829237ddf83db2906711471aa9465fe481/silnlp/common/linear_regression.py#L79)):*

There's already an outer loop `for _ in range(100):` in `perform_enhanced_linear_regression`, so it seems like one of either the outer or inner loop are unnecessary for bootstrap sampling unless I'm missing something.
</blockquote></details>

I believe this is correct.  Bootstrap sampling performs a series of resamplings of the data to get possible values for a parameter (the slope and intercept here).  Each resampling run iterates through the size of the data set, and I have it set up to do 100 resampling runs.

mshannon-sil (commented):
Thanks! For some reason though I'm not seeing the new commit even when I refresh github. Did you push it to the branch?

> @mshannon-sil made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/1103#-:-OwT3OHq6OXoaSkWwTWw:b-ohxk71)* status: all files reviewed, 2 unresolved discussions (waiting on benjaminking and mshannon-sil).

___
*[`silnlp/common/linear_regression.py` line 79 at r1](https://reviewable.io/reviews/sillsdev/silnlp/1103#-Ow4E7Qa6xrEruCsdw5O:-OwT-Mbt1cFKNCi75822:b4zqwv7) ([raw file](https://github.com/sillsdev/silnlp/blob/ecaf42829237ddf83db2906711471aa9465fe481/silnlp/common/linear_regression.py#L79)):*

I believe this is correct.  Bootstrap sampling performs a series of resamplings of the data to get possible values for a parameter (the slope and intercept here).  Each resampling run iterates through the size of the data set, and I have it set up to do 100 resampling runs.
</blockquote></details>

Right. But since you specify `size=n` for `np.random.choice`, isn't that already drawing n samples in just that line, thus iterating through the size of the dataset without needing to use a loop? Which would make the total samples `100*num_samples*n` rather than `100*n`. If we're intentionally trying to make the size of sampled_x and sampled_y to be `num_samples*n` rather than just `n`, then that's fine, just want to be clear.

benjaminking (commented):
Maybe I didn't push it before.  Sorry about that. It should be visible now.

> @benjaminking made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/1103#-:-OwV4-oH7xbXidmCGZN-:brp3v0)* status: 0 of 2 files reviewed, 2 unresolved discussions (waiting on mshannon-sil).

___
*[`silnlp/common/linear_regression.py` line 79 at r1](https://reviewable.io/reviews/sillsdev/silnlp/1103#-Ow4E7Qa6xrEruCsdw5O:-OwV3tcI3CfwtbpAdQMb:bfpfnat) ([raw file](https://github.com/sillsdev/silnlp/blob/ecaf42829237ddf83db2906711471aa9465fe481/silnlp/common/linear_regression.py#L79)):*

Right. But since you specify `size=n` for `np.random.choice`, isn't that already drawing n samples in just that line, thus iterating through the size of the dataset without needing to use a loop? Which would make the total samples `100*num_samples*n` rather than `100*n`. If we're intentionally trying to make the size of sampled_x and sampled_y to be `num_samples*n` rather than just `n`, then that's fine, just want to be clear.
</blockquote></details>

Ah, I see it now.  This was definitely resulting in the sampled sets being too large.  It has been corrected now.

mshannon-sil (approved):
:lgtm:

> @mshannon-sil reviewed 2 files and all commit messages, made 1 comment, and resolved 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/1103#-:-OwZ0i3k-qz_bFSzPfs0:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on benjaminking).

PR #1110, opened 2026-07-07
Claude suggested improvement to rclone_setup.sh

davidbaines:
This fix was entirely written by Claude. It seems to have resolved the issue I was having where by a local /root/M folder was preventing rclone from mounting because there was already data in the mountpoint. 

It says that it first checks whether rclone is running, and if it is doesn't try to run it again. 
What I noticed and didn't realise was that rclone seems to survive vscode terminal being closed. I had thought that VScode terminal that kicked off the rclone script needed to stay open in order for M to be accessible. However I can close VScode and still access data on M/

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/1110)

### DIFF
--- a/rclone_setup.sh
+++ b/rclone_setup.sh
@@ -1,32 +1,61 @@
 #!/bin/bash
-curl https://rclone.org/install.sh | bash
+
+BUCKET_TYPE=$1
+
+# Pick the mountpoint / remote for the requested bucket type.
+case "$BUCKET_TYPE" in
+    minio)
+        MOUNTPOINT="/root/M"
+        REMOTE="miniosilnlp:nlp-research"
+        ;;
+    backblaze)
+        MOUNTPOINT="/root/B"
+        REMOTE="b2silnlp:silnlp"
+        ;;
+    *)
+        echo "rclone_setup.sh: unknown bucket type '$BUCKET_TYPE' (expected 'minio' or 'backblaze')" >&2
+        return 1 2>/dev/null || exit 1
+        ;;
+esac
+
+# If the mountpoint is already a live, responsive rclone mount, there is nothing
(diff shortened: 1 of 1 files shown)


benjaminking (approved):
:lgtm:

But Matthew Beech should also review this.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/1110#-:-Owx4hod3SzwJdheOBT0:b-inmkg9)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on TaperChipmunk32).

rclone_setup.sh
```diff
@@ -1,32 +1,61 @@
 #!/bin/bash
-curl https://rclone.org/install.sh | bash
```
TaperChipmunk32, line 2: This line should be added back. It ensures that the Docker container has the latest rclone version.

TaperChipmunk32 (changes requested):
Could you also change the line endings for this file back to `LF`, instead of `CFLF`. If you are not sure how to do that, I can push a commit to change it as well. This will prevent syntax errors when running in Docker.

rclone_setup.sh
TaperChipmunk32: Since the two changes were so small, I decided to go ahead and push a commit myself to fix them.

TaperChipmunk32 (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/1110#-:-OwxoUec3Upf1_BjQNyZ:bnfp4nl)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on benjaminking and davidbaines).

PR #1111, opened 2026-07-07
Remove unused libraries from `pyproject.toml`

benjaminking:
This PR removes several libraries from SILNLP that appear to be unused.  Some of them are related to functionality we recently removed, but for others of them, it's not clear why they were included in the first place.  I have run the full experiment pipeline and verified that it completed successfully with these libraries removed.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/1111)


mshannon-sil (changes requested):
> @mshannon-sil reviewed 2 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/1111)* status: all files reviewed, 1 unresolved discussion (waiting on benjaminking).

___
*[`pyproject.toml` line 81 at r1](https://reviewable.io/reviews/sillsdev/silnlp/1111#-OwyZMOyAjmkcKjg9VA2:-OwyZMOyAjmkcKjg9VA3:b-v58xip) ([raw file](https://github.com/sillsdev/silnlp/blob/407cba2eca0d2ea84f0334f2df6dc51c41c215c5/pyproject.toml#L81)):*
> ```TOML
> accelerate = "^1.0.0"
> transformers = "^4.48"
> optimum = "^1.16.0"
> ```

The reason we had optimum is because I was experimenting with bettertransformer, which requires that library. However, we determined that we can't use it because it doesn't support attention masks for padding. It would probably be good to just remove all the code related to bettertransformer if we're getting rid of optimum.

benjaminking (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/1111)* status: 2 of 3 files reviewed, 1 unresolved discussion (waiting on mshannon-sil).

___
*[`pyproject.toml` line 81 at r1](https://reviewable.io/reviews/sillsdev/silnlp/1111#-OwyZMOyAjmkcKjg9VA2:-OyF8BlK4UHAz4Owf-Tw:b-896fix) ([raw file](https://github.com/sillsdev/silnlp/blob/407cba2eca0d2ea84f0334f2df6dc51c41c215c5/pyproject.toml#L81)):*

The reason we had optimum is because I was experimenting with bettertransformer, which requires that library. However, we determined that we can't use it because it doesn't support attention masks for padding. It would probably be good to just remove all the code related to bettertransformer if we're getting rid of optimum.
</blockquote></details>

Done.

mshannon-sil (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/1111#-:-OyFMui06ovglRPQiPHp:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on benjaminking).

PR #1115, opened 2026-07-09
Increase number of samples used for linear regression

benjaminking:
The PR increases the number of samples used for bootstrap sampling in the new linear regression method.  This should improve the stability of the regression results across different runs for a negligible runtime cost.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/1115)

### DIFF
--- a/silnlp/common/linear_regression.py
+++ b/silnlp/common/linear_regression.py
@@ -109,7 +109,7 @@ def perform_enhanced_linear_regression(x: List[float], y: List[float]) -> Linear
     weighting_scheme = InverseDensityPointWeightingScheme()
 
     bootstrap_results = []
-    for _ in range(100):
+    for _ in range(1000):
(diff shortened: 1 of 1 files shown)


mshannon-sil (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/1115#-:-Ox6I3ZNAb3iRS3Zle6z:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on benjaminking).

mshannon-sil:
I'm not sure that the randomness of the bootstrap method is actually the issue anymore. I've just rerun the linear regression calculation three times in a row, renaming the file each time so it doesn't get overwritten, and I got the same values each time.

benjaminking:
When I run this via `quality_estimation.py`, I do still get slightly different results between runs, though with the increased number of samples, the values are very close between multiple runs.

Issue #1114, opened 2026-07-09
Move the creation of linregress.json to the test step


mshannon-sil:
We should move the creation of linregress.json to the test step for a few reasons. First, all the information needed to create linregress.json (the chrf3 and confidence values) are in the test step. Second, if doing a two step process where one first holds out data to get the linear regression values and then runs a model with the full dataset, it doesn't make sense to require the first step to do translation and quality estimation. Third, if quality estimation is run multiple times, right now that means that the linear regression coefficients are calculated over and over. This is not only inefficient, but could also result in changes to the coefficients due to random sampling, which can interfere with comparisons across experiments that are supposed to be reusing the coefficients.

PR #1117, opened 2026-07-09
move creation of linregress.json to test step

mshannon-sil:
This PR moves the creation of linregress.json to the test step, as discussed in issue #1114. This way, it's only created once rather than each time quality estimation is run, and the `estimate_quality` method takes in a path to the linregress.json file rather than a path to the verse test scores.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/1117)


benjaminking (changes requested):
Good work identifying this change.  It seems like an improvement, not only for the user, but also for the architecture -- it makes the SILNLP a lot cleaner.

> @benjaminking reviewed 5 files and all commit messages, and made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/1117#-:-Ox7S7ZJ87Nzny_OIUf8:bys7jcd)* status: all files reviewed, 1 unresolved discussion (waiting on mshannon-sil).

___
*[`silnlp/nmt/experiment.py` line 138 at r1](https://reviewable.io/reviews/sillsdev/silnlp/1117#-Ox7P0KhAE3lpkqhhS3r:-Ox7P0KhAE3lpkqhhS3s:bu97byk) ([raw file](https://github.com/sillsdev/silnlp/blob/ec1453ee48a9ee48b40a7b201fc2978223d34545/silnlp/nmt/experiment.py#L138)):*
> ```Python
>                     self.produce_multiple_translations,
>                     self.save_confidences,
>                     bool(quality_estimation),
> ```

While we're doing this, are we able to combine the quality estimation and the linear regression parameters?  For example, can we only pass the linear regression path when we want to perform quality estimation?  And then only perform quality estimation in `translate` when a valid linear regression path is passed in?

mshannon-sil:
*[`silnlp/nmt/experiment.py` line 138 at r1](https://reviewable.io/reviews/sillsdev/silnlp/1117#-Ox7P0KhAE3lpkqhhS3r:-Ox7ZREEBYngh26L8S8p:b-reaf1g) ([raw file](https://github.com/sillsdev/silnlp/blob/ec1453ee48a9ee48b40a7b201fc2978223d34545/silnlp/nmt/experiment.py#L138)):*

While we're doing this, are we able to combine the quality estimation and the linear regression parameters?  For example, can we only pass the linear regression path when we want to perform quality estimation?  And then only perform quality estimation in `translate` when a valid linear regression path is passed in?
</blockquote></details>

Yeah, I think they should be combinable. Do you think we should keep them as separate flags in the command so that the user can choose to not specify the linear regression path if they don't have to (i.e., the linregress.json is located in the same experiment folder)? And just later in the code use the linregress.json file not being None as the flag for quality estimation?

benjaminking (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/1117)* status: 2 of 5 files reviewed, 1 unresolved discussion (waiting on mshannon-sil).

___
*[`silnlp/nmt/experiment.py` line 138 at r1](https://reviewable.io/reviews/sillsdev/silnlp/1117#-Ox7P0KhAE3lpkqhhS3r:-OxBD8Cy6yvG2b5vMq4d:b-d6j275) ([raw file](https://github.com/sillsdev/silnlp/blob/ec1453ee48a9ee48b40a7b201fc2978223d34545/silnlp/nmt/experiment.py#L138)):*

Yeah, I think they should be combinable. Do you think we should keep them as separate flags in the command so that the user can choose to not specify the linear regression path if they don't have to (i.e., the linregress.json is located in the same experiment folder)? And just later in the code use the linregress.json file not being None as the flag for quality estimation?
</blockquote></details>

Yes, that's probably fine to keep them as separate options so that a power user has the option to use different linear regression parameters.  We'll probably want the description of the linregress parameter to make it clear that it should not be included in standard usage.

mshannon-sil (commented):
> @mshannon-sil made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/1117)* status: 1 of 5 files reviewed, 1 unresolved discussion (waiting on benjaminking).

___
*[`silnlp/nmt/experiment.py` line 138 at r1](https://reviewable.io/reviews/sillsdev/silnlp/1117#-Ox7P0KhAE3lpkqhhS3r:-OxRf1m25PAfHp1RJVTk:b-896fix) ([raw file](https://github.com/sillsdev/silnlp/blob/ec1453ee48a9ee48b40a7b201fc2978223d34545/silnlp/nmt/experiment.py#L138)):*

Yes, that's probably fine to keep them as separate options so that a power user has the option to use different linear regression parameters.  We'll probably want the description of the linregress parameter to make it clear that it should not be included in standard usage.
</blockquote></details>

Done.

benjaminking (approved):
:lgtm:

> @benjaminking reviewed 4 files and all commit messages, made 1 comment, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/1117#-:-OxWXLf-8E2Y04_i1wwd:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on mshannon-sil).

PR #1123, opened 2026-07-23
update devcontainer for ease of use

mshannon-sil:
I updated the devcontainer to make it pick the correct python interpreter from the beginning and ensure the formatting extensions also automatically work from the beginning. There was a relatively new vscode python default extension that was overriding the desired functionality, so I needed to disable that for this to work. I also added claude code to the list of extensions, so that it doesn't need to be manually installed every time a devcontainer is recreated.

@TaperChipmunk32 can you double check that it works properly for you as well when you rebuild your devcontainer from scratch?

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/1123)

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -53,6 +53,19 @@
 			// Set *default* container specific settings.json values on container create.
 			"settings": {
 				"python.defaultInterpreterPath": "/workspaces/silnlp/.venv/bin/python",
+				"python.useEnvironmentsExtension": false,
+				"isort.interpreter": [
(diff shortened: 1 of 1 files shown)


TaperChipmunk32 (approved):
This seems to be working without any issues for me.
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/1123#-:-OyEX6am6sBhC9jkONah:b-4w89f4)* status: 0 of 1 files reviewed, all discussions resolved.

PR #1127, opened 2026-07-31
Update the Create onboarding experiments script and tests.

davidbaines:
Previous changes were to prevent Back Translations being used as drafting sources or as primary reference sources. The test used was simply looking for projects that have data in fewer books than the target project. That proved to be far to strict of a limit and excluded related projects from being a primary source. 

After a few rounds of changes the script now relies more on the specifications given by the user and gives warnings when the drafting source is missing 25% or more of the verses in the book (or chapters) being drafted.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/1127)


TaperChipmunk32 (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/1127#-:-OzMaeP38oM_GvFDbvga:bnfp4nl)* status: 0 of 2 files reviewed, all discussions resolved.

davidbaines:
I think I've complicated things a little bit here. I made more changes to the code before I merged in the previous set of changes. These latest changes use a namesOverride.json file so that the code uses names that the EITL team usually use for naming folders on M:.  We use "PNG" instead of "Papua New Guinea". Having the equivalents in a separate file avoids having to add the changes everywhere in the languageFamilies.json file.

PR #1128, opened 2026-08-04
Update linear regression and compute labels from pchrF3

benjaminking:
This PR updates the linear regression calculation to match the most recent method we've been targeting.  It also calculates usability labels directly from the projected chrF3 scores, instead of having usability scores as an intermediate.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/1128)

### DIFF
--- a/silnlp/common/linear_regression.py
+++ b/silnlp/common/linear_regression.py
@@ -114,12 +114,11 @@ class AverageLinearRegressionResultAggregator(LinearRegressionResultAggregator):
 
 def perform_enhanced_linear_regression(x: List[float], y: List[float]) -> LinearRegressionResult:
     sampler = BootstrapSampler(x, y)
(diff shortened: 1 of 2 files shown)


mshannon-sil (approved):
Sad to see the numeric usability scores go, but makes sense not to build a further metric on top of the chrf3 projection when the projection has so much variability itself. This will make it much easier to maintain and adapt this code as well.

LGTM assuming you've checked that the linear regression output now matches what we expect.

> @mshannon-sil reviewed 2 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/1128#-:-OzD-jyy0Obib8FdZa8j:bqz724t)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on benjaminking).

PR #1134, opened 2026-08-12
Hf upgrade

mshannon-sil:
This PR upgrades transformers to version 5.14, upgrades torch to 2.6, and upgrades a few other related libraries as well. Here's a list of key changes made due to the transformers upgrade:

- Some keys were renamed in 5.x as follows: RENAMED_CONFIG_KEYS = {
    "train": {"group_by_length": "train_sampling_strategy"},
    "eval": {"include_inputs_for_metrics": "include_for_metrics"},
    "params": {"warmup_ratio": "warmup_steps"},
}
- logging_dir arg deprecated in favor of TENSORBOARD_LOGGING_DIR env variable

- PreTrainedTokenizer type removed, replaced with PythonBackend and PreTrainedTokenizerBase types

- Changes related to 5.x removing the slow/fast tokenizer differentiation in typing

- special_tokens_map: Now only stores named special token attributes (e.g., bos_token, eos_token). The extra_special_tokens arg is used for additional special tokens (formerly additional_special_tokens). all_special_tokens includes both named and extra tokens.

- Rewrite batch_prepare_for_model due to tokenizer.prepare_for_model() no longer existing for fast tokenizers

- group_beam_search removed, so diverse_beam_search is no longer a possible multiple_translations_method without substantial rework, and I replaced the spot where it was called with a warning

- TranslationPipeline removed in 5.x, I replaced SilTranslationPipeline with SilTranslator class

- generation_config for model.generate is deprecated

- _get_train_sampler() now takes in a train_dataset arg

- apply_chat_template returns BatchEncoding by default rather than input_ids unless return_dict is set to False now

- self.args.auto_find_batch_size now needs to be set to true for automatic batch sizes

- report_to now needs to be set to all explicitly rather than be the default

Unrelated to upgrade: removed overriding save method in seq2seq_config.py, which was used for better transformers and lora, both of which have been removed from seq2seq_config.py already

I tested that experiment.py works properly for seq2seq and llm models, replicating chrf3 scores on a test project for both. It does seem like with the upgrades, seq2seq may be using slightly more memory as the auto_grad_acc ended up with a batch size of 8 rather than 16 for the same test project. We may want to look for ways to save some memory somehow if this is consistent behavior.

There's also a few minor changes to improve how formatters work in the devcontainer.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/1134)


TaperChipmunk32:
I tried running an experiment on this branch with Gemma4 and got the following error message: `from torch._C import *  # noqa: F403
ImportError: libcudnn.so.9: cannot open shared object file: No such file or directory`

mshannon-sil:
@TaperChipmunk32 I forgot to mention that this needs a new docker container for it to run properly. I created a local docker container called `silnlp:torch26` on both aqua and cheetah. So if you want to run it yourself, replace `ghcr.io/sillsdev/silnlp:latest` with `silnlp:torch26` in clearml_connection.py. As soon as this branch is merged in, then I'll update the regular silnlp docker image.

benjaminking (changes requested):
> @benjaminking reviewed 10 files and all commit messages, and made 12 comments.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/1134)* status: all files reviewed, 12 unresolved discussions (waiting on mshannon-sil).

___
*[`silnlp/nmt/llm_config.py` line 104 at r1](https://reviewable.io/reviews/sillsdev/silnlp/1134#-OzvhqcW8jrcKqPvGwHK:-OzvhqcW8jrcKqPvGwHL:bb0ls0u) ([raw file](https://github.com/sillsdev/silnlp/blob/ed88d21de2db81a912b25ab4a622cf44f0c99b85/silnlp/nmt/llm_config.py#L104)):*
> ```Python
>         "max_grad_norm",
>         "optim",
>         "warmup_ratio",
> ```

Should this be removed to match `seq2seq_config.py`?
___
*[`silnlp/nmt/llm_config.py` line 155 at r1](https://reviewable.io/reviews/sillsdev/silnlp/1134#-OzvgYBIBCejm66lPseX:-OzvgYBJ5JhiwcY99BR9:bhas0pc) ([raw file](https://github.com/sillsdev/silnlp/blob/ed88d21de2db81a912b25ab4a622cf44f0c99b85/silnlp/nmt/llm_config.py#L155)):*
> ```Python
> 
> 
> def apply_chat_template(
> ```

I feel like factoring this out into a function increases complexity, since it's a one-line function that replaces one line of code with another.
___
*[`silnlp/nmt/llm_config.py` line 633 at r1](https://reviewable.io/reviews/sillsdev/silnlp/1134#-Ozw1tMu5qbEXOrgW9V2:-Ozw1tMv-O41vNeOfg3f:bdbj1p7) ([raw file](https://github.com/sillsdev/silnlp/blob/ed88d21de2db81a912b25ab4a622cf44f0c99b85/silnlp/nmt/llm_config.py#L633)):*
> ```Python
>     ):
>         if self._auto_grad_acc:
>             (args if args is not None else self.args).auto_find_batch_size = True
> ```

Why would we want this to be set on `args` instead of `self.args`?
___
*[`silnlp/nmt/seq2seq_config.py` line 17 at r1](https://reviewable.io/reviews/sillsdev/silnlp/1134#-Ozw2VGy4fdD3D4MTcrD:-Ozw2VGy4fdD3D4MTcrE:b-5mmdeh) ([raw file](https://github.com/sillsdev/silnlp/blob/ed88d21de2db81a912b25ab4a622cf44f0c99b85/silnlp/nmt/seq2seq_config.py#L17)):*
> ```Python
> import numpy as np
> import pandas as pd
> import safetensors.torch
> ```

Is anything using `safetensors.torch` now that you removed `_save`?
___
*[`silnlp/nmt/seq2seq_config.py` line 233 at r1](https://reviewable.io/reviews/sillsdev/silnlp/1134#-Ozvn77N6yAwoLEAfwB_:-Ozvn77N6yAwoLEAfwBa:bw750vm) ([raw file](https://github.com/sillsdev/silnlp/blob/ed88d21de2db81a912b25ab4a622cf44f0c99b85/silnlp/nmt/seq2seq_config.py#L233)):*
> ```Python
>     # Huggingface does not follow its own type hints with this function and expects Dict[str, List[str]]
>     tokenizer.add_special_tokens(
>         {"additional_special_tokens": [lang_code]},  # pyright: ignore[reportArgumentType]
> ```

It looks like `additional_special_tokens` is deprecated and should be `extra_special_tokens` now.
___
*[`silnlp/nmt/seq2seq_config.py` line 698 at r1](https://reviewable.io/reviews/sillsdev/silnlp/1134#-OzvlDw24ndiJdczehIM:-OzvlDw24ndiJdczehIN:bal5vwk) ([raw file](https://github.com/sillsdev/silnlp/blob/ed88d21de2db81a912b25ab4a622cf44f0c99b85/silnlp/nmt/seq2seq_config.py#L698)):*
> ```Python
>                 if self.model_prefix == "facebook/nllb-200":
>                     self._tokenizer = NllbTokenizer.from_pretrained(str(self.exp_dir), token=False)
>                     self._tokenizer = convert_slow_tokenizer(self._tokenizer)
> ```

Do we still need to convert between slow and fast tokenizers with transformers 5?
___
*[`silnlp/nmt/seq2seq_config.py` line 706 at r1](https://reviewable.io/reviews/sillsdev/silnlp/1134#-OzvnQaSCka3qU26Miq3:-OzvnQaSCka3qU26Miq4:b-oh6jfk) ([raw file](https://github.com/sillsdev/silnlp/blob/ed88d21de2db81a912b25ab4a622cf44f0c99b85/silnlp/nmt/seq2seq_config.py#L706)):*
> ```Python
>                     self._tokenizer = T5Tokenizer(tokenizer_object=self._tokenizer)
>                     self._tokenizer.add_special_tokens(
>                         {"additional_special_tokens": ["<s>"]}, replace_extra_special_tokens=False
> ```

Same issue here with`additional_special_tokens` -> `extra_special_tokens`.
___
*[`silnlp/nmt/seq2seq_config.py` line 1677 at r1](https://reviewable.io/reviews/sillsdev/silnlp/1134#-Ozvtaj03S1lyNxryaPG:-Ozvtaj03S1lyNxryaPH:bf0h930) ([raw file](https://github.com/sillsdev/silnlp/blob/ed88d21de2db81a912b25ab4a622cf44f0c99b85/silnlp/nmt/seq2seq_config.py#L1677)):*
> ```Python
>         generate_kwargs["min_length"] = generate_kwargs.get("min_length", config.min_length)
>         generate_kwargs["max_length"] = generate_kwargs.get("max_length", config.max_length)
>         if input_length > 0.9 * generate_kwargs["max_length"]:
> ```

It looks like this was a Huggingface warning that no longer appears because `TranslationPipeline` has been replaced, but I'm not sure that we need to move the warning into our code.
___
*[`silnlp/nmt/seq2seq_config.py` line 1684 at r1](https://reviewable.io/reviews/sillsdev/silnlp/1134#-OzvzbuJ1F-_4SjhX_Al:-OzvzbuK4P07NcRVYRIv:bqoa57x) ([raw file](https://github.com/sillsdev/silnlp/blob/ed88d21de2db81a912b25ab4a622cf44f0c99b85/silnlp/nmt/seq2seq_config.py#L1684)):*
> ```Python
>                 generate_kwargs["max_length"],
>             )
>         output = self.model.generate(
> ```

My read of the transformers 5 changes is that `generate` does still expect a `generation_config` argument, but that it's no longer stored in the model parameters.  (https://github.com/huggingface/transformers/blob/main/MIGRATION_GUIDE_V5.md#generate) Can you double check on that?
___
*[`silnlp/nmt/seq2seq_config.py` line 1754 at r1](https://reviewable.io/reviews/sillsdev/silnlp/1134#-OzvvHik8EK4JsFqvqzL:-OzvvHik8EK4JsFqvqzM:b-rxbs3y) ([raw file](https://github.com/sillsdev/silnlp/blob/ed88d21de2db81a912b25ab4a622cf44f0c99b85/silnlp/nmt/seq2seq_config.py#L1754)):*
> ```Python
>                     {
>                         "translation_token_ids": sequence_ids.tolist(),
>                         "token_scores": token_scores[sentence_index][sequence_index],
> ```

Minor thing, but I think this would be cleaner if you also stored these values in variables before calling `records.append`.  It would make the lines shorter and easier to read.

(I like to think about it in terms of allowing the reader to focus on what is important to them -- if they don't care where `translation_text` is coming from, then it's easier to ignore how it's created and find what is important to them)
___
*[`silnlp/nmt/seq2seq_config.py` line 1770 at r1](https://reviewable.io/reviews/sillsdev/silnlp/1134#-Ozw38TOEEY5khJ_kktC:-Ozw38TOEEY5khJ_kktD:bekc9y0) ([raw file](https://github.com/sillsdev/silnlp/blob/ed88d21de2db81a912b25ab4a622cf44f0c99b85/silnlp/nmt/seq2seq_config.py#L1770)):*
> ```Python
>     def preprocess(self, sentences: List[Any]) -> BatchEncoding:
>         model_inputs = batch_prepare_for_model(self.tokenizer, sentences)
>         model_inputs = self.tokenizer.pad(model_inputs, padding=True, return_tensors="pt")
> ```

It looks like you moved the padding operation inside of `batch_prepare_for_model`, but forgot to delete it here, so the inputs are getting padded twice.
___
*[`tests/smoke_tests/experiments/test_experiment/config.yml` line 15 at r1](https://reviewable.io/reviews/sillsdev/silnlp/1134#-Ozw0UZjDT05GPBunimE:-Ozw0UZjDT05GPBunimF:b-gcb2fb) ([raw file](https://github.com/sillsdev/silnlp/blob/ed88d21de2db81a912b25ab4a622cf44f0c99b85/tests/smoke_tests/experiments/test_experiment/config.yml#L15)):*
> ```YAML
> eval:
>   per_device_eval_batch_size: 8
>   eval_strategy: "no"
> ```

What was the rationale behind these changes?

mshannon-sil (commented):
> @mshannon-sil made 11 comments.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/1134)* status: 9 of 10 files reviewed, 12 unresolved discussions (waiting on benjaminking).

___
*[`silnlp/nmt/llm_config.py` line 104 at r1](https://reviewable.io/reviews/sillsdev/silnlp/1134#-OzvhqcW8jrcKqPvGwHK:-OzwA8r02Xaxc2ga3Bbu:bgfcylk) ([raw file](https://github.com/sillsdev/silnlp/blob/ed88d21de2db81a912b25ab4a622cf44f0c99b85/silnlp/nmt/llm_config.py#L104)):*

Should this be removed to match `seq2seq_config.py`?
</blockquote></details>

Yes, missed this when I was fixing it for seq2seq_config.py
___
*[`silnlp/nmt/llm_config.py` line 155 at r1](https://reviewable.io/reviews/sillsdev/silnlp/1134#-OzvgYBIBCejm66lPseX:-OzwBMOm1AS7IaxRtbFn:b-9p6b5g) ([raw file](https://github.com/sillsdev/silnlp/blob/ed88d21de2db81a912b25ab4a622cf44f0c99b85/silnlp/nmt/llm_config.py#L155)):*

I feel like factoring this out into a function increases complexity, since it's a one-line function that replaces one line of code with another.
</blockquote></details>

Fair, I thought there was some typing reason why it should be extracted, but I'm not seeing the issue anymore so I'll make the change.
___
*[`silnlp/nmt/llm_config.py` line 633 at r1](https://reviewable.io/reviews/sillsdev/silnlp/1134#-Ozw1tMu5qbEXOrgW9V2:-OzwBVOr0KJiTrEFQTIU:b-cth84x) ([raw file](https://github.com/sillsdev/silnlp/blob/ed88d21de2db81a912b25ab4a622cf44f0c99b85/silnlp/nmt/llm_config.py#L633)):*

Why would we want this to be set on `args` instead of `self.args`?
</blockquote></details>

That's because args gets passed in to the inner_training_loop, not self.args, although I think in the internal library flow, args is set to self.args so it probably doesn't matter too much in practice. I modified the logic here to just remove the reference to self.args completely.
___
*[`silnlp/nmt/seq2seq_config.py` line 17 at r1](https://reviewable.io/reviews/sillsdev/silnlp/1134#-Ozw2VGy4fdD3D4MTcrD:-OzwIkS693mFmwe3TImj:b-kj8xw8) ([raw file](https://github.com/sillsdev/silnlp/blob/ed88d21de2db81a912b25ab4a622cf44f0c99b85/silnlp/nmt/seq2seq_config.py#L17)):*

Is anything using `safetensors.torch` now that you removed `_save`?
</blockquote></details>

No, removed.
___
*[`silnlp/nmt/seq2seq_config.py` line 233 at r1](https://reviewable.io/reviews/sillsdev/silnlp/1134#-Ozvn77N6yAwoLEAfwB_:-OzwQz4r-q8UvB5aAg7T:b-896fix) ([raw file](https://github.com/sillsdev/silnlp/blob/ed88d21de2db81a912b25ab4a622cf44f0c99b85/silnlp/nmt/seq2seq_config.py#L233)):*

It looks like `additional_special_tokens` is deprecated and should be `extra_special_tokens` now.
</blockquote></details>

Done.
___
*[`silnlp/nmt/seq2seq_config.py` line 698 at r1](https://reviewable.io/reviews/sillsdev/silnlp/1134#-OzvlDw24ndiJdczehIM:-OzwT--LEC-PrLGW98VB:b-pdwed3) ([raw file](https://github.com/sillsdev/silnlp/blob/ed88d21de2db81a912b25ab4a622cf44f0c99b85/silnlp/nmt/seq2seq_config.py#L698)):*

Do we still need to convert between slow and fast tokenizers with transformers 5?
</blockquote></details>

I believe so, the change that happened was the typing, not the distinction between slow and fast tokenizers altogether.
___
*[`silnlp/nmt/seq2seq_config.py` line 706 at r1](https://reviewable.io/reviews/sillsdev/silnlp/1134#-OzvnQaSCka3qU26Miq3:-OzwT5VN-EEFDo20N-9b:b-896fix) ([raw file](https://github.com/sillsdev/silnlp/blob/ed88d21de2db81a912b25ab4a622cf44f0c99b85/silnlp/nmt/seq2seq_config.py#L706)):*

Same issue here with`additional_special_tokens` -> `extra_special_tokens`.
</blockquote></details>

Done.
___
*[`silnlp/nmt/seq2seq_config.py` line 1677 at r1](https://reviewable.io/reviews/sillsdev/silnlp/1134#-Ozvtaj03S1lyNxryaPG:-OzwTWAW1RGGYYyh8Y4A:b-896fix) ([raw file](https://github.com/sillsdev/silnlp/blob/ed88d21de2db81a912b25ab4a622cf44f0c99b85/silnlp/nmt/seq2seq_config.py#L1677)):*

It looks like this was a Huggingface warning that no longer appears because `TranslationPipeline` has been replaced, but I'm not sure that we need to move the warning into our code.
</blockquote></details>

Done.
___
*[`silnlp/nmt/seq2seq_config.py` line 1684 at r1](https://reviewable.io/reviews/sillsdev/silnlp/1134#-OzvzbuJ1F-_4SjhX_Al:-OzwTmPfBV6X3VstDDTk:b1cngzj) ([raw file](https://github.com/sillsdev/silnlp/blob/ed88d21de2db81a912b25ab4a622cf44f0c99b85/silnlp/nmt/seq2seq_config.py#L1684)):*

My read of the transformers 5 changes is that `generate` does still expect a `generation_config` argument, but that it's no longer stored in the model parameters.  (https://github.com/huggingface/transformers/blob/main/MIGRATION_GUIDE_V5.md#generate) Can you double check on that?
</blockquote></details>

It seems that model.generate expects either the generation_config or the kwargs but not both. Their internal code indicates that a warning is produced if both are provided:
```
# It doesn't make sense to allow kwargs and `generation_config`, that should be mutually exclusive
        if generation_config_provided and set(kwargs.keys()) - set(model_kwargs.keys()):
            generation_kwargs = set(kwargs.keys()) - set(model_kwargs.keys())
            logger.warning_once(
                f"Passing `generation_config` together with generation-related "
                f"arguments=({generation_kwargs}) is deprecated and will be removed in future versions. "
                "Please pass either a `generation_config` object OR all generation "
                "parameters explicitly, but not both.",
            )
```
___
*[`silnlp/nmt/seq2seq_config.py` line 1770 at r1](https://reviewable.io/reviews/sillsdev/silnlp/1134#-Ozw38TOEEY5khJ_kktC:-Ozwwwld37ph2DWMWd-B:b-te6699) ([raw file](https://github.com/sillsdev/silnlp/blob/ed88d21de2db81a912b25ab4a622cf44f0c99b85/silnlp/nmt/seq2seq_config.py#L1770)):*

It looks like you moved the padding operation inside of `batch_prepare_for_model`, but forgot to delete it here, so the inputs are getting padded twice.
</blockquote></details>

In batch_prepare_for_model, padding is set to False. I'm only calling pad in there for a quick way to build the batch with the attention mask. So the only time actual padding happens is in this preprocess method right here.
___
*[`tests/smoke_tests/experiments/test_experiment/config.yml` line 15 at r1](https://reviewable.io/reviews/sillsdev/silnlp/1134#-Ozw0UZjDT05GPBunimE:-OzwwE4GCa102XPPFJdZ:b-iyxaku) ([raw file](https://github.com/sillsdev/silnlp/blob/ed88d21de2db81a912b25ab4a622cf44f0c99b85/tests/smoke_tests/experiments/test_experiment/config.yml#L15)):*

What was the rationale behind these changes?
</blockquote></details>

The test was trying to do an eval step at the end of training, causing it to fail, even though no eval step was required since it only runs 8 steps and eval_steps is set to 500 by default. So I just explicitly set it not to do eval, and the test works now.

mshannon-sil (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/1134)* status: 8 of 10 files reviewed, 11 unresolved discussions (waiting on benjaminking).

___
*[`silnlp/nmt/seq2seq_config.py` line 1754 at r1](https://reviewable.io/reviews/sillsdev/silnlp/1134#-OzvvHik8EK4JsFqvqzL:-Ozx-UfU7R4MnSTyebnT:b-896fix) ([raw file](https://github.com/sillsdev/silnlp/blob/ed88d21de2db81a912b25ab4a622cf44f0c99b85/silnlp/nmt/seq2seq_config.py#L1754)):*

Minor thing, but I think this would be cleaner if you also stored these values in variables before calling `records.append`.  It would make the lines shorter and easier to read.

(I like to think about it in terms of allowing the reader to focus on what is important to them -- if they don't care where `translation_text` is coming from, then it's easier to ignore how it's created and find what is important to them)
</blockquote></details>

Done.

benjaminking (changes requested):
> @benjaminking reviewed 2 files and all commit messages, made 5 comments, and resolved 8 discussions.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/1134)* status: all files reviewed, 3 unresolved discussions (waiting on mshannon-sil).

___
*[`silnlp/nmt/llm_config.py` line 633 at r1](https://reviewable.io/reviews/sillsdev/silnlp/1134#-Ozw1tMu5qbEXOrgW9V2:-P--O-Fl6egEjcpWFXOW:belii43) ([raw file](https://github.com/sillsdev/silnlp/blob/ed88d21de2db81a912b25ab4a622cf44f0c99b85/silnlp/nmt/llm_config.py#L633)):*

That's because args gets passed in to the inner_training_loop, not self.args, although I think in the internal library flow, args is set to self.args so it probably doesn't matter too much in practice. I modified the logic here to just remove the reference to self.args completely.
</blockquote></details>

Ah, okay.  I missed that.
___
*[`silnlp/nmt/seq2seq_config.py` line 698 at r1](https://reviewable.io/reviews/sillsdev/silnlp/1134#-OzvlDw24ndiJdczehIM:-P--qmQ9Es4mGilrkfQY:b-vkv4t7) ([raw file](https://github.com/sillsdev/silnlp/blob/ed88d21de2db81a912b25ab4a622cf44f0c99b85/silnlp/nmt/seq2seq_config.py#L698)):*

I believe so, the change that happened was the typing, not the distinction between slow and fast tokenizers altogether.
</blockquote></details>

It looks like the fast tokenizer is the default when you call `from_pretrained` (via [Tokenization in Transformers v5: Simpler, Clearer, and More Modular](https://huggingface.co/blog/tokenizers)), so I think we don't need to convert between the different classes anymore.
___
*[`silnlp/nmt/seq2seq_config.py` line 1684 at r1](https://reviewable.io/reviews/sillsdev/silnlp/1134#-OzvzbuJ1F-_4SjhX_Al:-P--rJmU7b6jmnS4zt6O:bd67pjr) ([raw file](https://github.com/sillsdev/silnlp/blob/ed88d21de2db81a912b25ab4a622cf44f0c99b85/silnlp/nmt/seq2seq_config.py#L1684)):*

It seems that model.generate expects either the generation_config or the kwargs but not both. Their internal code indicates that a warning is produced if both are provided:
```
# It doesn't make sense to allow kwargs and `generation_config`, that should be mutually exclusive
        if generation_config_provided and set(kwargs.keys()) - set(model_kwargs.keys()):
            generation_kwargs = set(kwargs.keys()) - set(model_kwargs.keys())
            logger.warning_once(
                f"Passing `generation_config` together with generation-related "
                f"arguments=({generation_kwargs}) is deprecated and will be removed in future versions. "
                "Please pass either a `generation_config` object OR all generation "
                "parameters explicitly, but not both.",
            )
```
</blockquote></details>

Ok, that makes sense.  The only thing I'd still want to double check is whether this logic from the old version is still needed in some form:

```
        # NLLB models incorrectly set max_length in the model config instead of the generation config
        if self._config.model_prefix == "facebook/nllb-200" and model.generation_config is not None:
            model.generation_config.max_length = model.config.max_length
            model.config.max_length = None
```
___
*[`silnlp/nmt/seq2seq_config.py` line 1770 at r1](https://reviewable.io/reviews/sillsdev/silnlp/1134#-Ozw38TOEEY5khJ_kktC:-P--rlev0rHUJDOl5F0D:bhfi10) ([raw file](https://github.com/sillsdev/silnlp/blob/ed88d21de2db81a912b25ab4a622cf44f0c99b85/silnlp/nmt/seq2seq_config.py#L1770)):*

In batch_prepare_for_model, padding is set to False. I'm only calling pad in there for a quick way to build the batch with the attention mask. So the only time actual padding happens is in this preprocess method right here.
</blockquote></details>

Ah, I missed that.  This should be fine then.
___
*[`tests/smoke_tests/experiments/test_experiment/config.yml` line 15 at r1](https://reviewable.io/reviews/sillsdev/silnlp/1134#-Ozw0UZjDT05GPBunimE:-P--u-CbA6b4lpA7nM0D:bh38qnp) ([raw file](https://github.com/sillsdev/silnlp/blob/ed88d21de2db81a912b25ab4a622cf44f0c99b85/tests/smoke_tests/experiments/test_experiment/config.yml#L15)):*

The test was trying to do an eval step at the end of training, causing it to fail, even though no eval step was required since it only runs 8 steps and eval_steps is set to 500 by default. So I just explicitly set it not to do eval, and the test works now.
</blockquote></details>

Was it something about the upgrade that caused it to do an eval step where it wasn't doing it before?

mshannon-sil (commented):
> @mshannon-sil made 3 comments.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/1134)* status: all files reviewed, 3 unresolved discussions (waiting on benjaminking).

___
*[`silnlp/nmt/seq2seq_config.py` line 698 at r1](https://reviewable.io/reviews/sillsdev/silnlp/1134#-OzvlDw24ndiJdczehIM:-P--vSs2BaUKHn0mKoZe:bcf6qv0) ([raw file](https://github.com/sillsdev/silnlp/blob/ed88d21de2db81a912b25ab4a622cf44f0c99b85/silnlp/nmt/seq2seq_config.py#L698)):*

It looks like the fast tokenizer is the default when you call `from_pretrained` (via [Tokenization in Transformers v5: Simpler, Clearer, and More Modular](https://huggingface.co/blog/tokenizers)), so I think we don't need to convert between the different classes anymore.
</blockquote></details>

Oh you're right, from_pretrained can work with just a sentence piece model now without doing a slow conversion first. Made that change now. Just had to add something to pass in the regular 200 NLLB lang code due to a quirk of initializing from just a sentencepiece file without the tokenizer.json. This isn't the normal path for initializing a tokenizer though so I don't expect it'll get run much outside of very experimental tokenizer work.
___
*[`silnlp/nmt/seq2seq_config.py` line 1684 at r1](https://reviewable.io/reviews/sillsdev/silnlp/1134#-OzvzbuJ1F-_4SjhX_Al:-P-1-p-o0uGafGz7fjW0:b-cjwmnc) ([raw file](https://github.com/sillsdev/silnlp/blob/ed88d21de2db81a912b25ab4a622cf44f0c99b85/silnlp/nmt/seq2seq_config.py#L1684)):*

Ok, that makes sense.  The only thing I'd still want to double check is whether this logic from the old version is still needed in some form:

```
        # NLLB models incorrectly set max_length in the model config instead of the generation config
        if self._config.model_prefix == "facebook/nllb-200" and model.generation_config is not None:
            model.generation_config.max_length = model.config.max_length
            model.config.max_length = None
```
</blockquote></details>

I think it shouldn't be needed, since I can see max_length=200 in the generation_config: https://huggingface.co/facebook/nllb-200-distilled-1.3B/blob/main/generation_config.json
___
*[`tests/smoke_tests/experiments/test_experiment/config.yml` line 15 at r1](https://reviewable.io/reviews/sillsdev/silnlp/1134#-Ozw0UZjDT05GPBunimE:-P-11V-Q-qqhXC8RbXMX:bmnxkcg) ([raw file](https://github.com/sillsdev/silnlp/blob/ed88d21de2db81a912b25ab4a622cf44f0c99b85/tests/smoke_tests/experiments/test_experiment/config.yml#L15)):*

Was it something about the upgrade that caused it to do an eval step where it wasn't doing it before?
</blockquote></details>

Yeah, if eval_strategy is set to "steps" (which is our default) then it now does an eval at the end of training even if it's not a multiple of the eval steps, which didn't happen in 4.x. However, maybe the better change if we don't want to include eval is justto set `type: train, test` under corpus_pairs in this config.yml for the test, since that's how we normally disable eval. Or else modify the test to work with the eval step. What do you think?

benjaminking (approved):
:lgtm:

> @benjaminking reviewed 1 file and all commit messages, made 2 comments, and resolved 3 discussions.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/1134#-:-P-7Cku5530TgWppTw7e:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on mshannon-sil).

___
*[`tests/smoke_tests/experiments/test_experiment/config.yml` line 15 at r1](https://reviewable.io/reviews/sillsdev/silnlp/1134#-Ozw0UZjDT05GPBunimE:-P-7Cgmi39T2K0juZll1:b7u0ljs) ([raw file](https://github.com/sillsdev/silnlp/blob/ed88d21de2db81a912b25ab4a622cf44f0c99b85/tests/smoke_tests/experiments/test_experiment/config.yml#L15)):*

Yeah, if eval_strategy is set to "steps" (which is our default) then it now does an eval at the end of training even if it's not a multiple of the eval steps, which didn't happen in 4.x. However, maybe the better change if we don't want to include eval is justto set `type: train, test` under corpus_pairs in this config.yml for the test, since that's how we normally disable eval. Or else modify the test to work with the eval step. What do you think?
</blockquote></details>

I think the changes you made are fine here.

PR #1140, opened 2026-08-18
Local and remote LLM support

benjaminking:
I'm creating a draft PR for this branch that adds support for mixing general instruction following examples into LLM fine-tuning.  I have downloaded a couple of instruction datasets to the bucket at `MT/instructions`.  Please let me know your thoughts on the approach that I've taken.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/1140)


benjaminking:
I have merged Damien's remote LLM branch into this branch, and it should now be ready for review.  Here is a quick rundown of what it contains:

* Support for general instruction following datasets, mixed in with translation instances during fine-tuning
* Fine-tuning on multi-turn instructions
* A script that allows you to chat with an LLM model (fine-tuned or off-the-shelf)
* More expressive prompts that allow for few-shot examples with in-context learning
* TF-IDF/BM25/semantic similarity as methods for example retrieval
* JSON/XML/arbitrary text formats for including examples in prompts
* The ability to fine-tune over a rotating list of prompts
* Remote LLM support via Lite LLM
* Token and cost usage tracking
* Retries for failed remote LLM queries

ddaspit (changes requested):
> @ddaspit reviewed 3 files and all commit messages, and made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/1140)* status: 3 of 25 files reviewed, 2 unresolved discussions (waiting on benjaminking, Enkidu93, mshannon-sil, and TaperChipmunk32).

___
*[`silnlp/nmt/example_retrieval.py` line 386 at r3](https://reviewable.io/reviews/sillsdev/silnlp/1140#-P1H-qwQAAePT7w51d_t:-P1H-qwQAAePT7w51d_u:b-io5q68) ([raw file](https://github.com/sillsdev/silnlp/blob/41d03c320735ed076b17219eebd16879e8ef761f/silnlp/nmt/example_retrieval.py#L386)):*
> ```Python
> 
>     def get_retriever(self) -> ExampleRetriever:
>         if not self._fitted:
> ```

This lazily initialized field is called inside of parallel code in `RemoteLLMConfig`, which can cause a race condition. Check out `RemoteLLMConfig._lock` to get an idea of how it handles race conditions like this.
___
*[`silnlp/nmt/local_llm_config.py` line 949 at r3](https://reviewable.io/reviews/sillsdev/silnlp/1140#-P1H3LuW9QUnfpsmqsGj:-P1H3LuW9QUnfpsmqsGk:b-xhnq9v) ([raw file](https://github.com/sillsdev/silnlp/blob/41d03c320735ed076b17219eebd16879e8ef761f/silnlp/nmt/local_llm_config.py#L949)):*
> ```Python
>         if training_args.max_steps and training_args.max_steps > 0:
>             return training_args.max_steps * effective_batch_size
>         return round(training_args.num_train_epochs * translation_size)
> ```

For epoch-based training, I think this should return the number of examples for one epoch, i.e. `translation_size * (1 + mix_ratio)`.

Enkidu93 (changes requested):
As we think about sharing more code between silnlp and machine.py, one thing that strikes me is that silnlp relies heavily on "config" classes to do a lot of the work whereas in machine.py, the logic is more separated across engine, model, build job, and file service classes. This isn't something we need to resolve now necessarily, but I think the machine.py organization structure may be easier to read, review, maintain, and test; and I don't think that switching to that kind of organization in silnlp would be as horrible as I previously thought. Just something to think about. 

> @Enkidu93 partially reviewed 25 files and all commit messages, and made 5 comments.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/1140#-:-P1VBD3x6j8sa7PL3U9H:bpf27nd)* status: all files reviewed, 6 unresolved discussions (waiting on benjaminking, mshannon-sil, and TaperChipmunk32).

___
*[`silnlp/nmt/example_retrieval.py` line 103 at r3](https://reviewable.io/reviews/sillsdev/silnlp/1140#-P1V97Os9yN8wE7Wokr9:-P1V97Os9yN8wE7WokrA:b-1ktlc9) ([raw file](https://github.com/sillsdev/silnlp/blob/41d03c320735ed076b17219eebd16879e8ef761f/silnlp/nmt/example_retrieval.py#L103)):*
> ```Python
>         directory.mkdir(parents=True, exist_ok=True)
>         with (directory / self._INDEX_FILENAME).open("wb") as file:
>             pickle.dump(self, file)
> ```

I wonder if pickle is the best solution here - both for space and long-term re-use. What exactly do we need to save and load here? One concern is that you could hit issues loading pkl files after dependencies or python versions change etc. Do we expect we might want to re-use these example retrievers after long periods of disuse. 
___
*[`silnlp/nmt/example_retrieval.py` line 243 at r3](https://reviewable.io/reviews/sillsdev/silnlp/1140#-P1VAtps1FRQrZ2762cX:-P1VAtps1FRQrZ2762cY:buzuk4r) ([raw file](https://github.com/sillsdev/silnlp/blob/41d03c320735ed076b17219eebd16879e8ef761f/silnlp/nmt/example_retrieval.py#L243)):*
> ```Python
> 
> 
> class ExampleRetrieverFactory:
> ```

Are there plans to add Mark's recent successful retrieval method using tfidf but trying to get better coverage rather than just the top most similar? Will that fit neatly into this scheme?
___
*[`pyproject.toml` line 107 at r3](https://reviewable.io/reviews/sillsdev/silnlp/1140#-P1Uy-Np9A88VCI1ZQWF:-P1Uy-Np9A88VCI1ZQWG:b-kv6nxd) ([raw file](https://github.com/sillsdev/silnlp/blob/41d03c320735ed076b17219eebd16879e8ef761f/pyproject.toml#L107)):*
> ```TOML
> [tool.poetry.extras]
> eflomal = ["eflomal"]
> llm = ["bitsandbytes", "sentence-transformers", "litellm", "rank-bm25"]
> ```

Are there plans to update the dockerfile to include the llm extra dependencies or what is the status of that?
___
*[`silnlp/nmt/config.py` line 475 at r3](https://reviewable.io/reviews/sillsdev/silnlp/1140#-P1UyxUzA0PErTliJwSC:-P1UyxUzA0PErTliJwSD:ba992kw) ([raw file](https://github.com/sillsdev/silnlp/blob/41d03c320735ed076b17219eebd16879e8ef761f/silnlp/nmt/config.py#L475)):*
> ```Python
>         self._delete_files("test.*.txt")
>         self._delete_files("dict.*.txt")
>         self._delete_files("instruction.*")
> ```

Is there a way to restructure this so that the instruction-LLM-support-specific calls are not in the generic config abstract base class? Could you, say, have a generic translation `write_data_files()` function and then have the `LlmConfig` call the base method as well as perform the LLM-specific functions? I'm not as familiar with the current conventions here in silnlp, so feel free to ignore if there's a reason that this needs to be in `Config`.

benjaminking (commented):
I have been thinking about this recently, and I would like to move toward having "config" classes in SILNLP perform less behavior and really just store configurations.  So that should very much be on the table as we refactor for machine.py.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/1140#-:-P1VYPIK6Cor_jIjEO_R:b-u651eb)* status: all files reviewed, 6 unresolved discussions (waiting on mshannon-sil and TaperChipmunk32).

mshannon-sil (changes requested):
> @mshannon-sil reviewed 9 files and all commit messages, and made 12 comments.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/1140)* status: all files reviewed, 17 unresolved discussions (waiting on benjaminking, Enkidu93, and TaperChipmunk32).

___
*[`pyproject.toml` line 107 at r3](https://reviewable.io/reviews/sillsdev/silnlp/1140#-P1Uy-Np9A88VCI1ZQWF:-P1aUw20-7d4Xto7C3OQ:b-da4d4e) ([raw file](https://github.com/sillsdev/silnlp/blob/41d03c320735ed076b17219eebd16879e8ef761f/pyproject.toml#L107)):*

Are there plans to update the dockerfile to include the llm extra dependencies or what is the status of that?
</blockquote></details>

Yes, we should release a new version of the docker container once this is merged. Let's make sure that the extras will get picked up by the docker container.
___
*[`scripts/chat_with_model.py` line 24 at r3](https://reviewable.io/reviews/sillsdev/silnlp/1140#-P1f_mosAqJdwgKFlbYr:-P1f_mosAqJdwgKFlbYs:b9ls9hd) ([raw file](https://github.com/sillsdev/silnlp/blob/41d03c320735ed076b17219eebd16879e8ef761f/scripts/chat_with_model.py#L24)):*
> ```Python
> 
> def main() -> None:
>     parser = argparse.ArgumentParser(description="Chat with a fine-tuned LLM checkpoint, or a base LLM")
> ```

Nit: In the description, I would include the word local somewhere so that it's clear this isn't for chatting with frontier models.
___
*[`scripts/chat_with_model.py` line 25 at r3](https://reviewable.io/reviews/sillsdev/silnlp/1140#-P1f_IyeEOo65ouFZmcU:-P1f_IyeEOo65ouFZmcV:bg1xla7) ([raw file](https://github.com/sillsdev/silnlp/blob/41d03c320735ed076b17219eebd16879e8ef761f/scripts/chat_with_model.py#L25)):*
> ```Python
> def main() -> None:
>     parser = argparse.ArgumentParser(description="Chat with a fine-tuned LLM checkpoint, or a base LLM")
>     parser.add_argument("target", help="Experiment name, or a bare HuggingFace model id/path")
> ```

Since target is often used in reference to a target language, I think we should choose a different word here to avoid confusion.
___
*[`scripts/chat_with_model.py` line 43 at r3](https://reviewable.io/reviews/sillsdev/silnlp/1140#-P1fnKzE47U_IZ5yCG6L:-P1fnKzE47U_IZ5yCG6M:b-4c2uec) ([raw file](https://github.com/sillsdev/silnlp/blob/41d03c320735ed076b17219eebd16879e8ef761f/scripts/chat_with_model.py#L43)):*
> ```Python
>         label = f"'{args.target}' (checkpoint={args.checkpoint})"
>     else:
>         # A config for a bare model name, with no experiment directory behind it: its model_dir
> ```

Nit: this comment is probably not necessary.
___
*[`silnlp/nmt/example_retrieval.py` line 275 at r3](https://reviewable.io/reviews/sillsdev/silnlp/1140#-P1lrrKo-HUZCRXzhd-L:-P1lrrKo-HUZCRXzhd-M:baww9) ([raw file](https://github.com/sillsdev/silnlp/blob/41d03c320735ed076b17219eebd16879e8ef761f/silnlp/nmt/example_retrieval.py#L275)):*
> ```Python
>     """Unlike JsonExampleFormatter/XmlExampleFormatter, does not escape the template output."""
> 
>     DEFAULT_TEMPLATE = "Source ({src_lang}): {source}\nTranslation ({trg_lang}): {target}\n\n"
> ```

Why does TextExampleFormatter include src_lang and trg_lang in its output, whereas json and xml do not?
___
*[`silnlp/nmt/example_retrieval.py` line 350 at r3](https://reviewable.io/reviews/sillsdev/silnlp/1140#-P1llG007NGFwkfOYQML:-P1llG007NGFwkfOYQMM:b-y1iwfb) ([raw file](https://github.com/sillsdev/silnlp/blob/41d03c320735ed076b17219eebd16879e8ef761f/silnlp/nmt/example_retrieval.py#L350)):*
> ```Python
>         return self._examples
> 
>     def _read_first_available_corpus(self) -> Optional[Tuple[List[str], List[str]]]:
> ```

What is the use case for creating logic to read the first available corpus? Are there scenarios where we expect an initial set of paths to not work, and we would want to use a fallback, rather than just throw an error?
___
*[`silnlp/nmt/example_retrieval.py` line 371 at r3](https://reviewable.io/reviews/sillsdev/silnlp/1140#-P1liv2-EvQLP-N6ahV0:-P1liv2000KRDd8OgU9_:b-6qgbna) ([raw file](https://github.com/sillsdev/silnlp/blob/41d03c320735ed076b17219eebd16879e8ef761f/silnlp/nmt/example_retrieval.py#L371)):*
> ```Python
> 
>     def select(self, query: str, k: int, pool_index: Optional[int] = None) -> List[Example]:
>         """Most relevant last, so the best examples sit nearest the source text. Excluding
> ```

Is this docstring completely accurate, if when every example fits, the examples are kept in corpus order?
___
*[`silnlp/nmt/llm_config.py` line 93 at r3](https://reviewable.io/reviews/sillsdev/silnlp/1140#-P1mKgu574-lnYYgcvpr:-P1mKgu574-lnYYgcvps:b-n3wks1) ([raw file](https://github.com/sillsdev/silnlp/blob/41d03c320735ed076b17219eebd16879e8ef761f/silnlp/nmt/llm_config.py#L93)):*
> ```Python
> class PromptTemplate:
>     _EXAMPLES_PLACEHOLDER = "{examples}"
>     _REQUIRED_FIELDS = ("system_message", "instruction_template", "example_format")
> ```

It looks like _REQUIRED_FIELDS should match the class's properties, but then should "example_format" actually be "formatter" here?
___
*[`silnlp/nmt/llm_config.py` line 169 at r3](https://reviewable.io/reviews/sillsdev/silnlp/1140#-P1mNvFH1BEAn70q4Dn1:-P1mNvFH1BEAn70q4Dn2:b-x79aug) ([raw file](https://github.com/sillsdev/silnlp/blob/41d03c320735ed076b17219eebd16879e8ef761f/silnlp/nmt/llm_config.py#L169)):*
> ```Python
>         self._templates = prompt_templates
> 
>     def is_empty(self) -> bool:
> ```

Nit: It might be more pythonic to just override __len__ here, up to you:

def __len__(self) -> int:
    return len(self._templates)
___
*[`silnlp/nmt/llm_config.py` line 191 at r3](https://reviewable.io/reviews/sillsdev/silnlp/1140#-P1mQMPI5GS0HQiOgSbS:-P1mQMPI5GS0HQiOgSbT:b303na5) ([raw file](https://github.com/sillsdev/silnlp/blob/41d03c320735ed076b17219eebd16879e8ef761f/silnlp/nmt/llm_config.py#L191)):*
> ```Python
>     def from_file(cls, prompt_file_path: Path) -> "PromptTemplateCollection":
>         if not prompt_file_path.is_file():
>             raise RuntimeError(f"The prompt template file {prompt_file_path} does not exist.")
> ```

I think you want a `FileNotFoundError` here.
___
*[`tests/unit_tests/test_example_retrieval.py` line 362 at r3](https://reviewable.io/reviews/sillsdev/silnlp/1140#-P1lutzT8e99-bGlejOB:-P1lutzT8e99-bGlejOC:bmt7krn) ([raw file](https://github.com/sillsdev/silnlp/blob/41d03c320735ed076b17219eebd16879e8ef761f/tests/unit_tests/test_example_retrieval.py#L362)):*
> ```Python
> 
> def test_tfidf_retriever_handles_a_corpus_with_no_word_tokens():
>     # TfidfVectorizer rejects an empty vocabulary outright, where bm25 returns nothing.
> ```

Should this be shown through a test of bm25, rather than communicated in a comment?
___
*[`tests/unit_tests/test_example_retrieval.py` line 368 at r3](https://reviewable.io/reviews/sillsdev/silnlp/1140#-P1luD89BNxUU9pzpJhN:-P1luD89BNxUU9pzpJhO:b-j1wgud) ([raw file](https://github.com/sillsdev/silnlp/blob/41d03c320735ed076b17219eebd16879e8ef761f/tests/unit_tests/test_example_retrieval.py#L368)):*
> ```Python
> 
> 
> def test_tfidf_retriever_ignores_sources_with_no_word_tokens():
> ```

Nit: I'm not sure this 100% demonstrates that it "ignores" sources with no word tokens, just that it didn't interfere with the match in this case.

mshannon-sil (changes requested):
> @mshannon-sil reviewed 2 files and made 4 comments.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/1140)* status: all files reviewed, 21 unresolved discussions (waiting on benjaminking, Enkidu93, and TaperChipmunk32).

___
*[`silnlp/nmt/llm_config.py` line 323 at r3](https://reviewable.io/reviews/sillsdev/silnlp/1140#-P1m_uuM7LiZfvzLT7SV:-P1m_uuM7LiZfvzLT7SW:b-rh0ap0) ([raw file](https://github.com/sillsdev/silnlp/blob/41d03c320735ed076b17219eebd16879e8ef761f/silnlp/nmt/llm_config.py#L323)):*
> ```Python
>         )
> 
>     def get_setting(self, key: str) -> Any:
> ```

Is this getter necessary? If it is, I think it should be used consistently in this class when accessing a key of self._settings.
___
*[`silnlp/nmt/llm_config.py` line 332 at r3](https://reviewable.io/reviews/sillsdev/silnlp/1140#-P1majG32xqTGvU21E8c:-P1majG32xqTGvU21E8d:b-98ym4e) ([raw file](https://github.com/sillsdev/silnlp/blob/41d03c320735ed076b17219eebd16879e8ef761f/silnlp/nmt/llm_config.py#L332)):*
> ```Python
>         self._settings["batch_instruction_template"] = template
> 
>     def resolve_defaults(self, defaults: PromptDefaults) -> None:
> ```

What's the motivation for this not being part of the init method?
___
*[`silnlp/nmt/llm_config.py` line 386 at r3](https://reviewable.io/reviews/sillsdev/silnlp/1140#-P1mcV9x-_uIzPAICNsm:-P1mcV9y8QKkdKaV3pES:b9iun3c) ([raw file](https://github.com/sillsdev/silnlp/blob/41d03c320735ed076b17219eebd16879e8ef761f/silnlp/nmt/llm_config.py#L386)):*
> ```Python
>             "data": {
>                 "mirror": False,
>                 "seed": 111,
> ```

Some of these values look the same as the seq2seq values. Should we move the ones they have in common up to config.py as part of a base class?
___
*[`silnlp/nmt/llm_config.py` line 460 at r3](https://reviewable.io/reviews/sillsdev/silnlp/1140#-P1md9bvDOD5_CEWnzR1:-P1md9bvDOD5_CEWnzR2:bi82bq2) ([raw file](https://github.com/sillsdev/silnlp/blob/41d03c320735ed076b17219eebd16879e8ef761f/silnlp/nmt/llm_config.py#L460)):*
> ```Python
> 
>     def create_tokenizer(self) -> Tokenizer:
>         # Data prep and test.py only tokenize and detokenize with this; both are raw text here.
> ```

I'm not sure I understand this comment. Could you remove it or make it clearer?

ddaspit (approved):
:lgtm:

> @ddaspit partially reviewed 23 files and all commit messages, made 1 comment, and resolved 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/1140#-:-P24bNUwE8HSBypUHs2K:bnfp4nl)* status: all files reviewed, 19 unresolved discussions (waiting on benjaminking, Enkidu93, and TaperChipmunk32).

benjaminking (commented):
> @benjaminking partially reviewed 6 files, made 21 comments, and resolved 7 discussions.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/1140)* status: 18 of 25 files reviewed, 12 unresolved discussions (waiting on ddaspit, Enkidu93, mshannon-sil, and TaperChipmunk32).

___
*[`pyproject.toml` line 107 at r3](https://reviewable.io/reviews/sillsdev/silnlp/1140#-P1Uy-Np9A88VCI1ZQWF:-P1eR09bC4Q-xIZ-9yvV:b-finqbp) ([raw file](https://github.com/sillsdev/silnlp/blob/41d03c320735ed076b17219eebd16879e8ef761f/pyproject.toml#L107)):*

Are there plans to update the dockerfile to include the llm extra dependencies or what is the status of that?
</blockquote></details>

Are you referring to the Docker image that runs on ClearML or the Dockerfile used to create dev containers?
___
*[`scripts/chat_with_model.py` line 24 at r3](https://reviewable.io/reviews/sillsdev/silnlp/1140#-P1f_mosAqJdwgKFlbYr:-P25LLO4B2N_pSX_T48j:b-896fix) ([raw file](https://github.com/sillsdev/silnlp/blob/41d03c320735ed076b17219eebd16879e8ef761f/scripts/chat_with_model.py#L24)):*

Nit: In the description, I would include the word local somewhere so that it's clear this isn't for chatting with frontier models.
</blockquote></details>

Done.
___
*[`scripts/chat_with_model.py` line 25 at r3](https://reviewable.io/reviews/sillsdev/silnlp/1140#-P1f_IyeEOo65ouFZmcU:-P25LMHJDevIrA-8nEtm:b-896fix) ([raw file](https://github.com/sillsdev/silnlp/blob/41d03c320735ed076b17219eebd16879e8ef761f/scripts/chat_with_model.py#L25)):*

Since target is often used in reference to a target language, I think we should choose a different word here to avoid confusion.
</blockquote></details>

Done.
___
*[`scripts/chat_with_model.py` line 43 at r3](https://reviewable.io/reviews/sillsdev/silnlp/1140#-P1fnKzE47U_IZ5yCG6L:-P25LN6GEBe8aJmZlWgp:b-896fix) ([raw file](https://github.com/sillsdev/silnlp/blob/41d03c320735ed076b17219eebd16879e8ef761f/scripts/chat_with_model.py#L43)):*

Nit: this comment is probably not necessary.
</blockquote></details>

Done.
___
*[`silnlp/nmt/config.py` line 475 at r3](https://reviewable.io/reviews/sillsdev/silnlp/1140#-P1UyxUzA0PErTliJwSC:-P1eTMwc8kM9LLFbGYjA:bkw9v66) ([raw file](https://github.com/sillsdev/silnlp/blob/41d03c320735ed076b17219eebd16879e8ef761f/silnlp/nmt/config.py#L475)):*

Is there a way to restructure this so that the instruction-LLM-support-specific calls are not in the generic config abstract base class? Could you, say, have a generic translation `write_data_files()` function and then have the `LlmConfig` call the base method as well as perform the LLM-specific functions? I'm not as familiar with the current conventions here in silnlp, so feel free to ignore if there's a reason that this needs to be in `Config`.
</blockquote></details>

Yes this is a good call.  Done.
___
*[`silnlp/nmt/example_retrieval.py` line 103 at r3](https://reviewable.io/reviews/sillsdev/silnlp/1140#-P1V97Os9yN8wE7Wokr9:-P1enpjwAQj-QWVQkVw4:b40qs1h) ([raw file](https://github.com/sillsdev/silnlp/blob/41d03c320735ed076b17219eebd16879e8ef761f/silnlp/nmt/example_retrieval.py#L103)):*

I wonder if pickle is the best solution here - both for space and long-term re-use. What exactly do we need to save and load here? One concern is that you could hit issues loading pkl files after dependencies or python versions change etc. Do we expect we might want to re-use these example retrievers after long periods of disuse.
</blockquote></details>

I did some profiling, and it seems like the benefit we'd get from caching is very small in the case of the lexical example retrievers, so those no longer cache.  There is still an advantage to caching the embedding similarity-based retriever's index (it takes 30 seconds to a couple minutes to construct), so I have switched to a different format (`.npy`) to store the matrix, which shouldn't have backward compatibility concerns.
___
*[`silnlp/nmt/example_retrieval.py` line 243 at r3](https://reviewable.io/reviews/sillsdev/silnlp/1140#-P1VAtps1FRQrZ2762cX:-P1escv00VJNAFSj2IzE:b-oxbh0c) ([raw file](https://github.com/sillsdev/silnlp/blob/41d03c320735ed076b17219eebd16879e8ef761f/silnlp/nmt/example_retrieval.py#L243)):*

Are there plans to add Mark's recent successful retrieval method using tfidf but trying to get better coverage rather than just the top most similar? Will that fit neatly into this scheme?
</blockquote></details>

I think it will be able to fit in well if we pass a reference to the list of already-selected examples in each retriever's constructor.  I think we can plan to do that in a separate PR though.
___
*[`silnlp/nmt/example_retrieval.py` line 275 at r3](https://reviewable.io/reviews/sillsdev/silnlp/1140#-P1lrrKo-HUZCRXzhd-L:-P25LgONDejDJ0KhrSAY:b-vvyau7) ([raw file](https://github.com/sillsdev/silnlp/blob/41d03c320735ed076b17219eebd16879e8ef761f/silnlp/nmt/example_retrieval.py#L275)):*

Why does TextExampleFormatter include src_lang and trg_lang in its output, whereas json and xml do not?
</blockquote></details>

It doesn't have to be this way, but this is what my thought process was: The text format is supposed to be very flexible, so I wanted to allow the use of the language names.  With the JSON and XML, however, I didn't want to introduce the possibility that language names might include reserved characters that weren't allowed for keys.  The JSON and XML also didn't have any natural semantics for which language was the source and which was the target (it's easier to include those semantics in a text format), so I thought it would be helpful to put that in the key names.  This could definitely be changed though.
___
*[`silnlp/nmt/example_retrieval.py` line 350 at r3](https://reviewable.io/reviews/sillsdev/silnlp/1140#-P1llG007NGFwkfOYQML:-P25YQlt-CdfrllyDIHO:btkhbzt) ([raw file](https://github.com/sillsdev/silnlp/blob/41d03c320735ed076b17219eebd16879e8ef761f/silnlp/nmt/example_retrieval.py#L350)):*

What is the use case for creating logic to read the first available corpus? Are there scenarios where we expect an initial set of paths to not work, and we would want to use a fallback, rather than just throw an error?
</blockquote></details>

This was a weird way to handle the logic for deciding between using the tokenized or detokenized versions of the training data.  It's been refactored now so that `ExamplePool` only ever sees a single corpus pair.
___
*[`silnlp/nmt/example_retrieval.py` line 371 at r3](https://reviewable.io/reviews/sillsdev/silnlp/1140#-P1liv2-EvQLP-N6ahV0:-P25Zjlf0LwVwse2vJvD:b-896fix) ([raw file](https://github.com/sillsdev/silnlp/blob/41d03c320735ed076b17219eebd16879e8ef761f/silnlp/nmt/example_retrieval.py#L371)):*

Is this docstring completely accurate, if when every example fits, the examples are kept in corpus order?
</blockquote></details>

Done.
___
*[`silnlp/nmt/example_retrieval.py` line 386 at r3](https://reviewable.io/reviews/sillsdev/silnlp/1140#-P1H-qwQAAePT7w51d_t:-P1f1gSGBpxVNQjGBiUU:b-896fix) ([raw file](https://github.com/sillsdev/silnlp/blob/41d03c320735ed076b17219eebd16879e8ef761f/silnlp/nmt/example_retrieval.py#L386)):*

This lazily initialized field is called inside of parallel code in `RemoteLLMConfig`, which can cause a race condition. Check out `RemoteLLMConfig._lock` to get an idea of how it handles race conditions like this.
</blockquote></details>

Done.
___
*[`silnlp/nmt/llm_config.py` line 93 at r3](https://reviewable.io/reviews/sillsdev/silnlp/1140#-P1mKgu574-lnYYgcvpr:-P25_UQf8SICuGd_aD6X:b-q5tkok) ([raw file](https://github.com/sillsdev/silnlp/blob/41d03c320735ed076b17219eebd16879e8ef761f/silnlp/nmt/llm_config.py#L93)):*

It looks like _REQUIRED_FIELDS should match the class's properties, but then should "example_format" actually be "formatter" here?
</blockquote></details>

In `from_json()`, we create `formatter` from `example_format`:

`formatter \= ExampleFormatterFactory.create(entry\["example\_format"\])`

So I think this is correct as written.
___
*[`silnlp/nmt/llm_config.py` line 169 at r3](https://reviewable.io/reviews/sillsdev/silnlp/1140#-P1mNvFH1BEAn70q4Dn1:-P25c-PG0E4GexVPmaN1:b-896fix) ([raw file](https://github.com/sillsdev/silnlp/blob/41d03c320735ed076b17219eebd16879e8ef761f/silnlp/nmt/llm_config.py#L169)):*

Nit: It might be more pythonic to just override __len__ here, up to you:

def __len__(self) -> int:
    return len(self._templates)
</blockquote></details>

Done.
___
*[`silnlp/nmt/llm_config.py` line 191 at r3](https://reviewable.io/reviews/sillsdev/silnlp/1140#-P1mQMPI5GS0HQiOgSbS:-P25cCWmBWdp9ITZdQXD:b-896fix) ([raw file](https://github.com/sillsdev/silnlp/blob/41d03c320735ed076b17219eebd16879e8ef761f/silnlp/nmt/llm_config.py#L191)):*

I think you want a `FileNotFoundError` here.
</blockquote></details>

Done.
___
*[`silnlp/nmt/llm_config.py` line 323 at r3](https://reviewable.io/reviews/sillsdev/silnlp/1140#-P1m_uuM7LiZfvzLT7SV:-P28UskD-ucU-7S9-lk5:b-93yva1) ([raw file](https://github.com/sillsdev/silnlp/blob/41d03c320735ed076b17219eebd16879e8ef761f/silnlp/nmt/llm_config.py#L323)):*

Is this getter necessary? If it is, I think it should be used consistently in this class when accessing a key of self._settings.
</blockquote></details>

I have removed this method.
___
*[`silnlp/nmt/llm_config.py` line 332 at r3](https://reviewable.io/reviews/sillsdev/silnlp/1140#-P1majG32xqTGvU21E8c:-P28XALyB-3c8Sp4g0ph:b4uhda1) ([raw file](https://github.com/sillsdev/silnlp/blob/41d03c320735ed076b17219eebd16879e8ef761f/silnlp/nmt/llm_config.py#L332)):*

What's the motivation for this not being part of the init method?
</blockquote></details>

It's because of the functionality that lets you specify a rotating set of prompts from a file.  If the prompt in the file has any field not set, it will fall back to the default, using this method.  I reorganized it a little bit to hopefully clarify this.
___
*[`silnlp/nmt/llm_config.py` line 386 at r3](https://reviewable.io/reviews/sillsdev/silnlp/1140#-P1mcV9x-_uIzPAICNsm:-P28bi4xDeNw23hOmFO_:b-74d4gu) ([raw file](https://github.com/sillsdev/silnlp/blob/41d03c320735ed076b17219eebd16879e8ef761f/silnlp/nmt/llm_config.py#L386)):*

Some of these values look the same as the seq2seq values. Should we move the ones they have in common up to config.py as part of a base class?
</blockquote></details>

You are right that it would make sense to do this, but I'm inclined to hold off on doing that for the moment, because it will involve making changes to `merge_dict()` in `Seq2SeqConfig`, and I don't want to make any changes there without thorough testing, which would further delay getting this merged.
___
*[`silnlp/nmt/llm_config.py` line 460 at r3](https://reviewable.io/reviews/sillsdev/silnlp/1140#-P1md9bvDOD5_CEWnzR1:-P28dPE4ECLykh-ba5oW:b-896fix) ([raw file](https://github.com/sillsdev/silnlp/blob/41d03c320735ed076b17219eebd16879e8ef761f/silnlp/nmt/llm_config.py#L460)):*

I'm not sure I understand this comment. Could you remove it or make it clearer?
</blockquote></details>

Done.
___
*[`silnlp/nmt/local_llm_config.py` line 949 at r3](https://reviewable.io/reviews/sillsdev/silnlp/1140#-P1H3LuW9QUnfpsmqsGj:-P1jz16f0b8SynkpyS_z:b-riwel4) ([raw file](https://github.com/sillsdev/silnlp/blob/41d03c320735ed076b17219eebd16879e8ef761f/silnlp/nmt/local_llm_config.py#L949)):*

For epoch-based training, I think this should return the number of examples for one epoch, i.e. `translation_size * (1 + mix_ratio)`.
</blockquote></details>

Done. This was also leading to an error if the user was using an "epoch" strategy where the same instructions instances would be used on each epoch instead of continuing to iterate through the instruction dataset.  That has also been corrected.
___
*[`tests/unit_tests/test_example_retrieval.py` line 362 at r3](https://reviewable.io/reviews/sillsdev/silnlp/1140#-P1lutzT8e99-bGlejOB:-P28fnXmE-UjBd1G4bg5:b-896fix) ([raw file](https://github.com/sillsdev/silnlp/blob/41d03c320735ed076b17219eebd16879e8ef761f/tests/unit_tests/test_example_retrieval.py#L362)):*

Should this be shown through a test of bm25, rather than communicated in a comment?
</blockquote></details>

Done.
___
*[`tests/unit_tests/test_example_retrieval.py` line 368 at r3](https://reviewable.io/reviews/sillsdev/silnlp/1140#-P1luD89BNxUU9pzpJhN:-P28iUAO9g1AZBlmdJMJ:bb54phh) ([raw file](https://github.com/sillsdev/silnlp/blob/41d03c320735ed076b17219eebd16879e8ef761f/tests/unit_tests/test_example_retrieval.py#L368)):*

Nit: I'm not sure this 100% demonstrates that it "ignores" sources with no word tokens, just that it didn't interfere with the match in this case.
</blockquote></details>

Good catch.  These tests are renamed and there's now a new test that covers what this one purported to cover.

mshannon-sil (changes requested):
> @mshannon-sil made 5 comments and resolved 8 discussions.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/1140)* status: 18 of 25 files reviewed, 6 unresolved discussions (waiting on benjaminking, ddaspit, Enkidu93, and TaperChipmunk32).

___
*[`silnlp/nmt/example_retrieval.py` line 275 at r3](https://reviewable.io/reviews/sillsdev/silnlp/1140#-P1lrrKo-HUZCRXzhd-L:-P29co9x8Oq3GovHA8Gp:b-p87v0u) ([raw file](https://github.com/sillsdev/silnlp/blob/41d03c320735ed076b17219eebd16879e8ef761f/silnlp/nmt/example_retrieval.py#L275)):*

It doesn't have to be this way, but this is what my thought process was: The text format is supposed to be very flexible, so I wanted to allow the use of the language names.  With the JSON and XML, however, I didn't want to introduce the possibility that language names might include reserved characters that weren't allowed for keys.  The JSON and XML also didn't have any natural semantics for which language was the source and which was the target (it's easier to include those semantics in a text format), so I thought it would be helpful to put that in the key names.  This could definitely be changed though.
</blockquote></details>

That makes sense, thanks.
___
*[`silnlp/nmt/llm_config.py` line 93 at r3](https://reviewable.io/reviews/sillsdev/silnlp/1140#-P1mKgu574-lnYYgcvpr:-P29eHId0uOIfDLyDT7b:b-htt1th) ([raw file](https://github.com/sillsdev/silnlp/blob/41d03c320735ed076b17219eebd16879e8ef761f/silnlp/nmt/llm_config.py#L93)):*

In `from_json()`, we create `formatter` from `example_format`:

`formatter \= ExampleFormatterFactory.create(entry\["example\_format"\])`

So I think this is correct as written.
</blockquote></details>

I see that now, thanks.
___
*[`silnlp/nmt/llm_config.py` line 386 at r3](https://reviewable.io/reviews/sillsdev/silnlp/1140#-P1mcV9x-_uIzPAICNsm:-P29ejrd7BP6ZFs75NJf:b-74iwvo) ([raw file](https://github.com/sillsdev/silnlp/blob/41d03c320735ed076b17219eebd16879e8ef761f/silnlp/nmt/llm_config.py#L386)):*

You are right that it would make sense to do this, but I'm inclined to hold off on doing that for the moment, because it will involve making changes to `merge_dict()` in `Seq2SeqConfig`, and I don't want to make any changes there without thorough testing, which would further delay getting this merged.
</blockquote></details>

Yeah that makes sense.
___
*[`silnlp/nmt/local_llm_config.py` line 411 at r3](https://reviewable.io/reviews/sillsdev/silnlp/1140#-P1q2EwIE0WvMtvDKjFH:-P1q2EwIE0WvMtvDKjFI:bdp3pv0) ([raw file](https://github.com/sillsdev/silnlp/blob/41d03c320735ed076b17219eebd16879e8ef761f/silnlp/nmt/local_llm_config.py#L411)):*
> ```Python
>     def get_instruction_mix_ratio(self) -> float:
>         ratio = float(self.train["instruction_data"]["mix_ratio"])
>         if ratio < 0:
> ```

Nit: If you're doing a check for 0, maybe do a check for a ratio greater than 1 as well.
___
*[`silnlp/nmt/local_llm_config.py` line 444 at r3](https://reviewable.io/reviews/sillsdev/silnlp/1140#-P1q44jqA6kWg6DdgeYD:-P1q44jrAVKE_tIoMB7k:bju2n7g) ([raw file](https://github.com/sillsdev/silnlp/blob/41d03c320735ed076b17219eebd16879e8ef761f/silnlp/nmt/local_llm_config.py#L444)):*
> ```Python
>                     )
>                 # Datasets mix evenly regardless of their original relative sizes; a dataset
>                 # smaller than its even share is used in full, without repetition (see split_corpus).
> ```

Should there be some kind of warning if dataset is smaller than its share size? It seems that if that happens, the train size becomes smaller than the user requested.

Enkidu93 (approved):
:lgtm:

> @Enkidu93 reviewed 11 files and all commit messages, made 4 comments, and resolved 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/1140#-:-P29njC79Spa4gfmS-il:bnfp4nl)* status: all files reviewed, 4 unresolved discussions (waiting on benjaminking, mshannon-sil, and TaperChipmunk32).

___
*[`pyproject.toml` line 107 at r3](https://reviewable.io/reviews/sillsdev/silnlp/1140#-P1Uy-Np9A88VCI1ZQWF:-P29mu4K5y92sLSTM5rP:b-mmj209) ([raw file](https://github.com/sillsdev/silnlp/blob/41d03c320735ed076b17219eebd16879e8ef761f/pyproject.toml#L107)):*

Are you referring to the Docker image that runs on ClearML or the Dockerfile used to create dev containers?
</blockquote></details>

Not sure if you are asking me or Matthew. He would know better. I don't use the dev container. I just figured this would need to be added to a poetry export or something maybe. 
___
*[`silnlp/nmt/example_retrieval.py` line 103 at r3](https://reviewable.io/reviews/sillsdev/silnlp/1140#-P1V97Os9yN8wE7Wokr9:-P29nLAF2zFOe5B8eHKx:bpal0n2) ([raw file](https://github.com/sillsdev/silnlp/blob/41d03c320735ed076b17219eebd16879e8ef761f/silnlp/nmt/example_retrieval.py#L103)):*

I did some profiling, and it seems like the benefit we'd get from caching is very small in the case of the lexical example retrievers, so those no longer cache.  There is still an advantage to caching the embedding similarity-based retriever's index (it takes 30 seconds to a couple minutes to construct), so I have switched to a different format (`.npy`) to store the matrix, which shouldn't have backward compatibility concerns.
</blockquote></details>

Great! Looks good.
___
*[`silnlp/nmt/example_retrieval.py` line 243 at r3](https://reviewable.io/reviews/sillsdev/silnlp/1140#-P1VAtps1FRQrZ2762cX:-P29nRXDDIqlZGOrvV6G:b-78i276) ([raw file](https://github.com/sillsdev/silnlp/blob/41d03c320735ed076b17219eebd16879e8ef761f/silnlp/nmt/example_retrieval.py#L243)):*

I think it will be able to fit in well if we pass a reference to the list of already-selected examples in each retriever's constructor.  I think we can plan to do that in a separate PR though.
</blockquote></details>

OK, sounds good. Just trying to avoid future work if there's a way to make room for that now.

mshannon-sil (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/1140)* status: all files reviewed, 3 unresolved discussions (waiting on benjaminking, Enkidu93, and TaperChipmunk32).

___
*[`pyproject.toml` line 107 at r3](https://reviewable.io/reviews/sillsdev/silnlp/1140#-P1Uy-Np9A88VCI1ZQWF:-P2F3tLm7pX4vpqlntmN:b-rtousv) ([raw file](https://github.com/sillsdev/silnlp/blob/41d03c320735ed076b17219eebd16879e8ef761f/pyproject.toml#L107)):*

Not sure if you are asking me or Matthew. He would know better. I don't use the dev container. I just figured this would need to be added to a poetry export or something maybe.
</blockquote></details>

Yes, we should add it to the devcontainer. At the end of devcontainer.json, there's a poststart command that installs poetry requirements. We should modify that to install all extras. Same thing with the equivalent line in the top level dockerfile.

mshannon-sil (changes requested):
> @mshannon-sil made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/1140)* status: all files reviewed, 4 unresolved discussions (waiting on benjaminking, Enkidu93, and TaperChipmunk32).

___
*[`silnlp/nmt/local_llm_config.py` line 883 at r5](https://reviewable.io/reviews/sillsdev/silnlp/1140#-P2IlykL7GsrgerM3TWW:-P2IlykL7GsrgerM3TWX:bnyiz6w) ([raw file](https://github.com/sillsdev/silnlp/blob/a9cd25fcef9036d4a164dfd1543d5dd77283d575/silnlp/nmt/local_llm_config.py#L883)):*
> ```Python
>         def encode_completion(prompt_ids: List[int], target: str) -> dict:
>             completion_ids = tokenizer(target, add_special_tokens=False)["input_ids"] + [eos_token_id]
>             input_ids = (prompt_ids + completion_ids)[:max_seq_length]
> ```

Devin caught a bug here that happens if the prompt_ids alone are longer than max sequence length, especially when doing few-shot examples. If that happens, the loss will be NaN and could poision the gradient. The recommended fixes were either truncating the beginning of the prompt and retain space for at least the completion and EOS token, or else dropping this datapoint. We should probably log if this happens too.

PR #1142, opened 2026-08-24
Add standalone scorer for scoring inference produced outside silnlp

laura-burdick-sil:
Adds silnlp-nmt-score, a self-contained CLI that reproduces the test stage's exact sacrebleu calls/parameters so LLM drafts and other non-silnlp inference can be scored with numbers directly comparable to silnlp's own scores-*.csv. Verified against all 17 standard_research_configs experiments and against silnlp.nmt.test's own scoring functions via a drift test.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/1142)


benjaminking (changes requested):
I think this script is trying to do way too much, including some things that don't really make sense to do (such as including confidence as a scorer, since we'll realistically never have confidence files that weren't created by SILNLP).  Some of this would be fixed by factoring out the evaluation code and having this and `test.py` both share that code.  But I'd recommend going through the arguments that this script accepts and filter it to ones that we're likely to use.

> @benjaminking reviewed 3 files and all commit messages, and made 3 comments.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/1142#-:-P-pGx3I53wPq7Syl8XO:b4ol56k)* status: all files reviewed, 2 unresolved discussions (waiting on laura-burdick-sil and mshannon-sil).

___
*[`silnlp/nmt/score.py` line 4 at r1](https://reviewable.io/reviews/sillsdev/silnlp/1142#-P-pE9kn9Vm7c9sbm-pV:-P-pE9kn9Vm7c9sbm-pW:b-exdafo) ([raw file](https://github.com/sillsdev/silnlp/blob/c2fad26faccb15c54183927c72a0a92ae9c04e92/silnlp/nmt/score.py#L4)):*
> ```Python
> parameters as ``silnlp.nmt.test`` does during the ``test`` stage.
> 
> This module is deliberately self-contained: it duplicates the sacrebleu calls in
> ```

Would it be possible to factor out the scoring code and have both `test.py` and this file use that implementation?  I'm concerned that we would be creating two implementations that we'd have to keep in sync.
___
*[`silnlp/nmt/score.py` line 11 at r1](https://reviewable.io/reviews/sillsdev/silnlp/1142#-P-pD4OT7KxZUYHrkRtp:-P-pD4OU6xZRhPgCypro:bh6mstq) ([raw file](https://github.com/sillsdev/silnlp/blob/c2fad26faccb15c54183927c72a0a92ae9c04e92/silnlp/nmt/score.py#L11)):*
> ```Python
> change a metric call here, change the drift test too - that test is what keeps this file honest.
> 
> Known, deliberate differences from ``test.py``:
> ```

Can you shorten up these comments a bit?  Claude can be a bit verbose at times.

mshannon-sil:
I second the idea to factor out the scoring code so that it can be used both by test.py and this file.

Also @benjaminking, I know in the past we were trying to avoid having comments unless necessary, to encourage writing self documenting code. That was in the pre-Claude world. Do we still want to avoid comments, or do we need to update our thinking on the frequency of comments in the repo?

ddaspit (commented):
I think comments are fine. The problem is more that Claude tends to write a whole novel in the code comments. I always tell Claude to review the comments to ensure that they are concise and that they are only used for complex or non-obvious code.

> @ddaspit made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/1142#-:-P-t2un96Upt-t3VGN4w:b-a6zigb)* status: all files reviewed, 2 unresolved discussions (waiting on laura-burdick-sil and mshannon-sil).

benjaminking (commented):
My personal philosophy is to only include comments for code that is non-intuitive or non-obvious and to rely on clean code that documents itself through good organization and good naming.

But as far as Claude's comments are concerned, they're definitely more than most people would read.  But given that SILNLP is not a particularly clean code base to start with, I want to err on the side of making coding agents useful and time-saving by not being overly strict.  All that being said, if I weren't recommending a refactoring, I'd probably have more comments on the comments.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/1142#-:-P-t743N8NVPRhwOrv3e:b2zkbf2)* status: all files reviewed, 2 unresolved discussions (waiting on laura-burdick-sil and mshannon-sil).

laura-burdick-sil (commented):
> @laura-burdick-sil made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/1142)* status: 1 of 5 files reviewed, 2 unresolved discussions (waiting on benjaminking and mshannon-sil).

___
*[`silnlp/nmt/score.py` line 4 at r1](https://reviewable.io/reviews/sillsdev/silnlp/1142#-P-pE9kn9Vm7c9sbm-pV:-P-tqi4VDGbX9F9YLx4S:b-896fix) ([raw file](https://github.com/sillsdev/silnlp/blob/c2fad26faccb15c54183927c72a0a92ae9c04e92/silnlp/nmt/score.py#L4)):*

Would it be possible to factor out the scoring code and have both `test.py` and this file use that implementation?  I'm concerned that we would be creating two implementations that we'd have to keep in sync.
</blockquote></details>

Done.
___
*[`silnlp/nmt/score.py` line 11 at r1](https://reviewable.io/reviews/sillsdev/silnlp/1142#-P-pD4OT7KxZUYHrkRtp:-P-tql153iAbSbsom4yD:b-896fix) ([raw file](https://github.com/sillsdev/silnlp/blob/c2fad26faccb15c54183927c72a0a92ae9c04e92/silnlp/nmt/score.py#L11)):*

Can you shorten up these comments a bit?  Claude can be a bit verbose at times.
</blockquote></details>

Done.

benjaminking (changes requested):
> @benjaminking reviewed 4 files and all commit messages, and made 5 comments.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/1142)* status: all files reviewed, 7 unresolved discussions (waiting on laura-burdick-sil and mshannon-sil).

___
*[`silnlp/nmt/score.py` line 26 at r2](https://reviewable.io/reviews/sillsdev/silnlp/1142#-P0WpdYH7ryubGfO6egE:-P0WpdYH7ryubGfO6egF:b55hrui) ([raw file](https://github.com/sillsdev/silnlp/blob/ac72191fd4fd980414557f6b04c2e60f6722de67/silnlp/nmt/score.py#L26)):*
> ```Python
> 
> 
> def check_version() -> None:
> ```

I'm not sure we need this.  Since this script is part of SILNLP, I think we can assume that users are running the same version of sacrebleu that is specified in `pyproject.toml`.
___
*[`silnlp/nmt/score.py` line 34 at r2](https://reviewable.io/reviews/sillsdev/silnlp/1142#-P0Wrmhm9Zf8TzBMTdvy:-P0Wrmhm9Zf8TzBMTdvz:b-4j5vf3) ([raw file](https://github.com/sillsdev/silnlp/blob/ac72191fd4fd980414557f6b04c2e60f6722de67/silnlp/nmt/score.py#L34)):*
> ```Python
> 
> 
> def write_verse_scores(
> ```

Can you use `PairScore.write()` instead of creating this method?
___
*[`silnlp/nmt/score.py` line 109 at r2](https://reviewable.io/reviews/sillsdev/silnlp/1142#-P0Wsm40CgMIa5Dm4q6N:-P0Wsm40CgMIa5Dm4q6O:b-4zdrcc) ([raw file](https://github.com/sillsdev/silnlp/blob/ac72191fd4fd980414557f6b04c2e60f6722de67/silnlp/nmt/score.py#L109)):*
> ```Python
>     )
>     parser.add_argument(
>         "--sacrebleu-tokenize",
> ```

Just to simplify things, I think you could drop this argument.  I don't think anyone ever uses a different tokenizer for sacrebleu.
___
*[`silnlp/nmt/scoring_metrics.py` line 91 at r2](https://reviewable.io/reviews/sillsdev/silnlp/1142#-P0WjzG8DT_nncGLMFWS:-P0WjzG8DT_nncGLMFWT:b-m76q35) ([raw file](https://github.com/sillsdev/silnlp/blob/ac72191fd4fd980414557f6b04c2e60f6722de67/silnlp/nmt/scoring_metrics.py#L91)):*
> ```Python
>     scorers: Set[str],
>     sacrebleu_tokenize: Optional[str] = None,
> ) -> Tuple[Optional[BLEUScore], Dict[str, float]]:
> ```

It could be good to create a class to represent a set of scores, since this is a construct we use over and over again.
___
*[`silnlp/nmt/scoring_metrics.py` line 241 at r2](https://reviewable.io/reviews/sillsdev/silnlp/1142#-P0WiwSv7NSQnMPLBjUi:-P0WiwSv7NSQnMPLBjUj:b-7m89s8) ([raw file](https://github.com/sillsdev/silnlp/blob/ac72191fd4fd980414557f6b04c2e60f6722de67/silnlp/nmt/scoring_metrics.py#L241)):*
> ```Python
>                 other_verse_scores["TER"] = ter_verse_score.score
> 
>         yield index, pred, sentences, bleu_verse_score, other_verse_scores
> ```

Can you create a class that represents the return value of this function?  When functions have very complex signatures like this one, it's easier to accidentally mis-assign the implicit return values, but using a class forces consumers to be explicit.

laura-burdick-sil (commented):
Thanks for the helpful feedback, Ben.

> @laura-burdick-sil made 6 comments.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/1142#-:-P1H67Di3QRglyYsR69C:bj0fsns)* status: 3 of 5 files reviewed, 7 unresolved discussions (waiting on benjaminking and mshannon-sil).

___
*[`silnlp/nmt/score.py` line 26 at r2](https://reviewable.io/reviews/sillsdev/silnlp/1142#-P0WpdYH7ryubGfO6egE:-P1H609H-wViEZhe5tcr:b-896fix) ([raw file](https://github.com/sillsdev/silnlp/blob/ac72191fd4fd980414557f6b04c2e60f6722de67/silnlp/nmt/score.py#L26)):*

I'm not sure we need this.  Since this script is part of SILNLP, I think we can assume that users are running the same version of sacrebleu that is specified in `pyproject.toml`.
</blockquote></details>

Done.
___
*[`silnlp/nmt/score.py` line 34 at r2](https://reviewable.io/reviews/sillsdev/silnlp/1142#-P0Wrmhm9Zf8TzBMTdvy:-P1H60e47hB2XYY_16TZ:b-896fix) ([raw file](https://github.com/sillsdev/silnlp/blob/ac72191fd4fd980414557f6b04c2e60f6722de67/silnlp/nmt/score.py#L34)):*

Can you use `PairScore.write()` instead of creating this method?
</blockquote></details>

Done.
___
*[`silnlp/nmt/score.py` line 109 at r2](https://reviewable.io/reviews/sillsdev/silnlp/1142#-P0Wsm40CgMIa5Dm4q6N:-P1H628t4s9uAOc8F17T:b-896fix) ([raw file](https://github.com/sillsdev/silnlp/blob/ac72191fd4fd980414557f6b04c2e60f6722de67/silnlp/nmt/score.py#L109)):*

Just to simplify things, I think you could drop this argument.  I don't think anyone ever uses a different tokenizer for sacrebleu.
</blockquote></details>

Done.
___
*[`silnlp/nmt/scoring_metrics.py` line 91 at r2](https://reviewable.io/reviews/sillsdev/silnlp/1142#-P0WjzG8DT_nncGLMFWS:-P1H64SQ0YvLtaHSIRxt:b-896fix) ([raw file](https://github.com/sillsdev/silnlp/blob/ac72191fd4fd980414557f6b04c2e60f6722de67/silnlp/nmt/scoring_metrics.py#L91)):*

It could be good to create a class to represent a set of scores, since this is a construct we use over and over again.
</blockquote></details>

Done.
___
*[`silnlp/nmt/scoring_metrics.py` line 241 at r2](https://reviewable.io/reviews/sillsdev/silnlp/1142#-P0WiwSv7NSQnMPLBjUi:-P1H65ONEtBYfGJvsPwW:b-896fix) ([raw file](https://github.com/sillsdev/silnlp/blob/ac72191fd4fd980414557f6b04c2e60f6722de67/silnlp/nmt/scoring_metrics.py#L241)):*

Can you create a class that represents the return value of this function?  When functions have very complex signatures like this one, it's easier to accidentally mis-assign the implicit return values, but using a class forces consumers to be explicit.
</blockquote></details>

Done.

PR #1150, opened 2026-09-11
Fix ValueError when reading conf file

TaperChipmunk32:
This PR fixes a bug that caused this error message during some test runs `ValueError: could not convert string to float: ''`. This error occurs when the verse text uses multiple lines, which interrupts the usual confidence file pattern of one line being sequence number and the verse text and the next line being the confidences.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/1150)


benjaminking (changes requested):
I'm wondering if a better way to address this might be to encode the newline characters as '\n' in the confidences file, because having a newline character in the text will also make it hard to view manually in other tools.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/1150#-:-P1V-KzM9skUd1YPoAU2:b-cgp04y)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on mshannon-sil).

PR #1154, opened 2026-09-18
Use new eflomal implementation

Enkidu93:
Also, add drafting training data to word alignment model training data. 

I've tested this on a real project using marker placement and noticed what I believe is comparable or better placement. This update will align the behavior in silnlp with what is in Serval/machine.py. At some point, I plan to evaluate more formally whether we ought to be training on the pretranslations at all.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/1154)


Enkidu93:
This is in Serval, so it'd be good to have it in silnlp as well - as you have time. Happy for an AI review even to get started on any problems.

benjaminking (changes requested):
Do you also need to modify `alignment/config.py` to point "eflomal" to the new Machine.py version?  It looks like it's still pointing to the old eflomal implementation.

> @benjaminking reviewed 6 files and all commit messages, and made 4 comments.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/silnlp/1154#-:-P2EgCskAxeMwZjxdp-p:bffralw)* status: all files reviewed, 3 unresolved discussions (waiting on ddaspit and Enkidu93).

___
*[`silnlp/common/postprocesser.py` line 337 at r1](https://reviewable.io/reviews/sillsdev/silnlp/1154#-P2EhXvMCZcwKJj4DYCy:-P2EhXvMCZcwKJj4DYCz:bjq0b5l) ([raw file](https://github.com/sillsdev/silnlp/blob/908786c9b43c6faa97c347ed6fa541e915155a90/silnlp/common/postprocesser.py#L337)):*
> ```Python
>         )
> 
>     def create_place_markers_postprocessor(self, training_corpus_pairs: List[CorpusPair] = []) -> PlaceMarkersPostprocessor:
> ```

It looks like `training_corpus_pairs` is never read.  This is also probably a situation where it would be better to have it be `Optional[List[CorpusPair]] = None`, due to the mutable argument issue.
___
*[`silnlp/common/translator.py` line 438 at r1](https://reviewable.io/reviews/sillsdev/silnlp/1154#-P2EaBbDBQh2h5j573ZD:-P2EaBbEBVbG952H0syo:b-bz618j) ([raw file](https://github.com/sillsdev/silnlp/blob/908786c9b43c6faa97c347ed6fa541e915155a90/silnlp/common/translator.py#L438)):*
> ```Python
>             postprocess_handler = PostprocessHandler(environment=self._environment)
>         for draft_index, translated_draft in enumerate(translated_text_rows.get_translated_drafts(), 1):
>             translated_text_rows.construct_postprocessing_rows_for_draft_index(postprocess_handler, draft_index,  list(load_corpus(experiment_config.exp_dir / experiment_config.train_src_detok_filename())),list(load_corpus(experiment_config.exp_dir / experiment_config.train_trg_detok_filename())) )
> ```

Could you inject these directories in these methods instead of injecting `Config`?  It would help reduce coupling if `Translator` didn't need to know about the experiment config.

Maybe you could wrap these in an object that only reads them once.  Right now, they're being read once per drafted book (times however many drafts we're producing).
___
*[`silnlp/nmt/postprocess.py` line 282 at r1](https://reviewable.io/reviews/sillsdev/silnlp/1154#-P2EfhAT7rm_KeffEgT5:-P2EfhAT7rm_KeffEgT6:b-sqa89a) ([raw file](https://github.com/sillsdev/silnlp/blob/908786c9b43c6faa97c347ed6fa541e915155a90/silnlp/nmt/postprocess.py#L282)):*
> ```Python
>                 environment=environment,
>                 train_src_filepath=config.exp_dir / config.train_src_detok_filename(),
>                 train_trg_filename=config.exp_dir / config.train_trg_detok_filename()
> ```

I believe this should be `train_trg_filepath`.

